PR #4, opened 2015-11-25
Webview

LeanderK:
Coded the web-view. There are some little bits left, but these are some small details.
Also added some test-data and a class to quickly test them on your computer.


LeanderK:
Yeah!

PR #12, opened 2016-01-17
Router

kelunik:
- Implements `Router`.
- Accepts `application/json` and `application/protobuf` payloads.
- Returns `application/json` or `application/protobuf` based on the `accept` header.

Other types are rejected with the correct HTTP status codes `415` respectively `406`.


LeanderK:
What about the hooks for the database? I thought the router provides them? Or ost it part of the event-system?

kelunik:
Database operations will be called in the `TemplateResource` handlers and similar TBD handlers.

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/api/Router.java
```diff
@@ -0,0 +1,127 @@
...
+        exception(BadRequestException.class, (exception, request, response) -> {
+            response.status(400);
+            response.type("application/json");
+            response.body(gson.toJson(new ErrorResponse("badRequest", exception.getMessage())));
+        });
+
+        exception(NotAcceptableException.class, (exception, request, response) -> {
+            response.status(406);
+            response.type("application/json");
+            response.body(gson.toJson(new ErrorResponse("notAcceptable", exception.getMessage())));
```
marcelhollerbach: I dont like to have one part of the api as json and the other part as protobuf. 
Why not just have protobuf ? For all ?

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/api/TemplateResource.java
```diff
@@ -0,0 +1,87 @@
...
+    public Template get(Request request, Response response) {
+        int id;
+
+        try {
+            id = Integer.parseInt(request.params(":id"));
+        } catch (NumberFormatException e) {
+            throw new BadRequestException(":id must be a valid integer.");
+        }
+
+        return Template.newBuilder().setId(id).setContent("{{TEST}}").build();
```
marcelhollerbach: Why merging the branch if there is placeholder code ? Why not directly connect to the db ?

marcelhollerbach:
Also, why naming the package "api" ? wouldnt router be somehow better?

kelunik:
Router is just a single class, but the package contains more than just a router.

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/api/TemplateResource.java
kelunik: Database operations do not exist yet. Well, partially for templates in 6e347052bbf0f5b1e6b653123eee24fe5c7d34ae.

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/api/Router.java
kelunik: There's support for both, except for error responses currently, right. Will write a special error responder.

kelunik: Errors are transformed the same way as every other response now: 2334a36be5d5700b0623f172ab00964f5f0c7832.

LeanderK:
but api is a very meaningless name, everything is an api. We also have internal database-apis, the event api etc. why not rest? why not zoidberg?

kelunik:
`rest` sounds good to me.

PR #20, opened 2016-01-20
Make empty lists return 200 instead of 404

kelunik:


LeanderK:
why does it return 202 instead of 404 now?

LeanderK:
it would be cool if you could wait for me if you change my code

kelunik:
Anything wrong with it?

LeanderK:
no, but there might be, also it's code in my package and i'd like to keep the control over it. Its not easy (especially in the database package) to keep track of all the changes been introduced all the the time.

PR #26, opened 2016-01-23
Crowdplatform2

marcelhollerbach:
crowdplatform basic shit

### DIFF
--- a/src/main/java/edu/kit/ipd/crowdcontrol/objectservice/crowdworking/Platform.java
+++ b/src/main/java/edu/kit/ipd/crowdcontrol/objectservice/crowdworking/Platform.java
@@ -19,7 +19,7 @@ public interface Platform {
      * if the Platform has his own worker identification the interface can be returned here
      * @return The value to indicate if it supports worker identification or not.
      */
-    Optional<Worker> getWorker();
+    Optional<WorkerIdentification> getWorker();
 
     /**
      * Get the name of this platform
--- a/src/main/java/edu/kit/ipd/crowdcontrol/objectservice/crowdworking/Platform.java
+++ b/src/main/java/edu/kit/ipd/crowdcontrol/objectservice/crowdworking/Platform.java
@@ -1,0 +1,58 @@
+package edu.kit.ipd.crowdcontrol.objectservice.crowdworking;
+
+import edu.kit.ipd.crowdcontrol.objectservice.proto.Experiment;
+
+import java.util.Optional;
+import java.util.concurrent.CompletableFuture;
+
+/**
+ * Created by marcel on 19.01.16.
+ */
+public interface Platform {
+    /**
+    * if the Platform has his own payment service the implementation can return not none.
+    * @return The value to indicate if it supports self paying or not.
+    */
+    Optional<Payment> getPayment();
+
+    /**
+     * if the Platform has his own worker identification the interface can be returned here
+     * @return The value to indicate if it supports worker identification or not.
+     */
+    Optional<Worker> getWorker();
+
+    /**
+     * Get the name of this platform
+     * @return A unique String
+     */
+    String getName();
+
+    /**
+     * Publish a passed experiment on the platform
+     * @param experiment the object which should be published
+     * @return This should return a unique string which is used to identify the published experiment later.
+     *         Will return None if the publishing failed
+     */
+    CompletableFuture<String> publishTask(Experiment experiment);
+
+    /**
+     * Unpublish the given id from the platform, after this call no worker should be able access the before published experiment.
+     * @param id The id of the before published experiment
+     * @return true on success, false on fail
+     */
+    CompletableFuture<Boolean> unpublishTask(String id);
+
+    /**
+     * Update the published task, with the given id, to the parameters of experiment
+     * @param id The id of the published Task
+     * @param experiment The experiment with the new parameters
+     * @return The new id
+     */
+    CompletableFuture<String> updateTask(String id, Experiment experiment);
+
+    /**
+     * Returns if calibration questions can be asked on this platform
+     * @return true or false
+     */
+    Boolean isCalibsAllowd();
+}
--- a/src/main/java/edu/kit/ipd/crowdcontrol/objectservice/crowdworking/Platform.java
+++ b/src/main/java/edu/kit/ipd/crowdcontrol/objectservice/crowdworking/Platform.java
@@ -54,5 +54,5 @@ public interface Platform {
      * Returns if calibration questions can be asked on this platform
      * @return true or false
      */
-    Boolean isCalibsAllowd();
+    Boolean isCalibrationAllowed();
 }
--- a/src/main/java/edu/kit/ipd/crowdcontrol/objectservice/crowdworking/Platform.java
+++ b/src/main/java/edu/kit/ipd/crowdcontrol/objectservice/crowdworking/Platform.java
@@ -31,14 +31,14 @@ public interface Platform {
      * Publish a passed experiment on the platform
      * @param experiment the object which should be published
      * @return This should return a unique string which is used to identify the published experiment later.
-     *         Will return None if the publishing failed
+     *         Or finish with a exception if the publishing failed.
      */
     CompletableFuture<String> publishTask(Experiment experiment);
 
     /**
      * Unpublish the given id from the platform, after this call no worker should be able access the before published experiment.
      * @param id The id of the before published experiment
-     * @return true on success, false on fail
+     * @return true on success, false or a exception if it failed
      */
     CompletableFuture<Boolean> unpublishTask(String id);
 
@@ -46,7 +46,7 @@ public interface Platform {
      * Update the published task, with the given id, to the parameters of experiment
      * @param id The id of the published Task
      * @param experiment The experiment with the new parameters
-     * @return The new id
+     * @return The new id or a Exception if the update failed.
      */
     CompletableFuture<String> updateTask(String id, Experiment experiment);
 
--- a/src/main/java/edu/kit/ipd/crowdcontrol/objectservice/crowdworking/PlatformManager.java
+++ b/src/main/java/edu/kit/ipd/crowdcontrol/objectservice/crowdworking/PlatformManager.java
@@ -190,8 +190,7 @@ public class PlatformManager {
      */
     public Optional<WorkerRecord> getWorker(String name, Map<String, String[]> params) {
         return identifyWorker(name, params).map(uid ->
-            workerOps.getAllWorkers().stream().
-                    filter(workerRecord1 -> workerRecord1.getIdentification().equals(uid)).findFirst().
+            workerOps.getWorker(name, uid).
                     orElseGet(() -> {
                         //create a new entry in the database
                         WorkerRecord record = new WorkerRecord(-1, uid, name, null);
--- a/src/main/java/edu/kit/ipd/crowdcontrol/objectservice/crowdworking/PlatformManager.java
+++ b/src/main/java/edu/kit/ipd/crowdcontrol/objectservice/crowdworking/PlatformManager.java
@@ -27,6 +27,7 @@ public class PlatformManager {
     private final Worker fallbackWorker;
     private final Payment fallbackPayment;
     private TasksOperations tasksOps;
+    private WorkerOperations workerOps;
 
     /**
      * Create a new manager for platforms. The known platforms in the database will be deleted,
@@ -115,19 +116,34 @@ public class PlatformManager {
      * @return None if the platform does not exist
      */
     public CompletableFuture<Boolean> publishTask(String name, Experiment experiment) throws TaskOperationException {
-        if (tasksOps.searchTask(name,experiment.getId()).isPresent())
-            throw new TaskOperationException("Experiment is already published!");
+
+        TaskRecord record = new TaskRecord();
+        record.setExperiment(experiment.getId());
+        record.setStatus(TaskStatus.running);
+        record.setCrowdPlatform(name);
+
+        if (tasksOps.createTask(record) == null)
+            throw new TaskOperationException("Task could not be created");
 
         return getPlatform(name)
                 .map(platform1 -> platform1.publishTask(experiment))
                 .orElseThrow(() -> new IllegalArgumentException("Experiment not found!"))
-                .thenApply(s -> {
-                    TaskRecord record = new TaskRecord();
-                    record.setExperiment(experiment.getId());
-                    record.setPlatformData(s);
-                    record.setStatus(TaskStatus.running);
-                    record.setCrowdPlatform(name);
-                    return tasksOps.createTask(record) != null;
+                .handle((s1, throwable) -> {
+                    if (s1 != null) {
+                        record.setCrowdPlatform(s1);
+                    } else {
+                        record.setStatus(TaskStatus.stopped);
+                    }
+                    if (!tasksOps.updateTask(record)) {
+                        if (s1 != null)
+                            try {
+                                unpublishTask(name, experiment).join();
+                            } catch (TaskOperationException e) {
+                                e.printStackTrace();
+                            }
+                        throw new IllegalStateException("Updating record for published task failed");
+                    }
+                    return true;
                 });
     }
 
@@ -183,6 +199,37 @@ public class PlatformManager {
         return getWorker(name).identifyWorker(params);
     }
 
+    /**
+     * Get a worker if he exists
+     * @param name Name of the platform
+     * @param params Params passed by the platform
+     * @return A worker if one is found
+     * @throws UnknownWorkerException if the platform does not identify a worker
+     */
+    public Optional<WorkerRecord> getWorker(String name, Map<String ,String[]> params) throws UnknownWorkerException {
+        String uid = identifyWorker(name, params);
+
+        return workerOps.getWorker(name, uid);
+    }
+
+    /**
+     * A Worker in every case
+     * @param name Name of the Platform
+     * @param params Params passed by the platform
+     * @return A worker which is identified by the params
+     * @throws UnknownWorkerException if the platform does not identify a worker
+     */
+    public WorkerRecord createWorker(String name, Map<String, String[]> params) throws UnknownWorkerException {
+        String uid = identifyWorker(name, params);
+
+        return workerOps.getWorker(name, uid).orElseGet(() -> {
+            WorkerRecord wr = new WorkerRecord();
+            wr.setIdentification(uid);
+            wr.setPlatform(name);
+
+            return workerOps.createWorker(wr);
+        });
+    }
     /**
      * Pay a worker
      * @param name The name of the platform
--- a/src/main/java/edu/kit/ipd/crowdcontrol/objectservice/crowdworking/PlatformManager.java
+++ b/src/main/java/edu/kit/ipd/crowdcontrol/objectservice/crowdworking/PlatformManager.java
@@ -27,7 +27,6 @@ public class PlatformManager {
     private final Worker fallbackWorker;
     private final Payment fallbackPayment;
     private TasksOperations tasksOps;
-    private WorkerOperations workerOps;
 
     /**
      * Create a new manager for platforms. The known platforms in the database will be deleted,
@@ -41,13 +40,11 @@ public class PlatformManager {
      * @param fallbackPayment handler which is called if a platform does not support payment
      * @param tasksOps Used for the task operations on the database
      * @param platformOps Used for the platform operations on the database
-     * @param workerOps Used for the worker operations on the database
      */
     public PlatformManager(List<Platform> crowdPlatforms, Worker fallbackWorker,
                            Payment fallbackPayment, TasksOperations tasksOps,
-                           PlatformOperations platformOps, WorkerOperations workerOps) {
+                           PlatformOperations platformOps) {
         this.tasksOps = tasksOps;
-        this.workerOps = workerOps;
         this.fallbackWorker = fallbackWorker;
         this.fallbackPayment = fallbackPayment;
 
@@ -90,8 +87,10 @@ public class PlatformManager {
      * @param name The name of the platform
      * @return The interface used to identify a worker
      */
-    public Optional<Worker> getWorker(String name) {
-        return getPlatform(name).map(platform -> platform.getWorker().orElse(fallbackWorker));
+    public Worker getWorker(String name) {
+        return getPlatform(name)
+                .orElseThrow(() -> new IllegalArgumentException("Platform not found"))
+                .getWorker().orElse(fallbackWorker);
     }
 
     /**
@@ -101,8 +100,10 @@ public class PlatformManager {
      * @param name The name of the platform to use
      * @return The interface used for payment
      */
-    public Optional<Payment> getPlatformPayment(String name) {
-        return getPlatform(name).map(platform -> platform.getPayment().orElse(fallbackPayment));
+    public Payment getPlatformPayment(String name) {
+        return getPlatform(name)
+                .orElseThrow(() -> new IllegalArgumentException("Platform not found"))
+                .getPayment().orElse(fallbackPayment);
     }
 
     /**
@@ -178,29 +179,18 @@ public class PlatformManager {
      * @param params Params passed by the platform
      * @return A String if the platform exists
      */
-    public Optional<String> identifyWorker(String name, Map<String, String[]> params) {
-        return getWorker(name).flatMap(worker -> worker.identifyWorker(params));
+    public String identifyWorker(String name, Map<String, String[]> params) throws UnknownWorkerException {
+        return getWorker(name).identifyWorker(params);
     }
 
     /**
-     * Get a worker record from the database which is associated with the given parameters
-     * @param name Name of the platform
-     * @param params Params which were passed by the platform to the workerservice
-     * @return A WorkerRecord if one is found
+     * Pay a worker
+     * @param name The name of the platform
+     * @param worker Worker to pay
+     * @param amount The amount of money
+     * @return A completable future which returns the success of the call
      */
-    public Optional<WorkerRecord> getWorker(String name, Map<String, String[]> params) {
-        return identifyWorker(name, params).map(uid ->
-            workerOps.getWorker(name, uid).
-                    orElseGet(() -> {
-                        //create a new entry in the database
-                        WorkerRecord record = new WorkerRecord(-1, uid, name, null);
-                        record = workerOps.createWorker(record);
-                        return record;
-                    })
-        );
-    }
-
-    public Optional<CompletableFuture<Boolean>> payWorker(String name, Worker worker, int amount) {
-        return getPlatformPayment(name).map(payment -> payment.payWorker(worker, amount));
+    public CompletableFuture<Boolean> payWorker(String name, Worker worker, int amount) {
+        return getPlatformPayment(name).payWorker(worker, amount);
     }
 }
--- a/src/main/java/edu/kit/ipd/crowdcontrol/objectservice/crowdworking/PlatformManager.java
+++ b/src/main/java/edu/kit/ipd/crowdcontrol/objectservice/crowdworking/PlatformManager.java
@@ -113,22 +113,21 @@ public class PlatformManager {
      * @param experiment The experiment to publish
      * @return None if the platform does not exist
      */
-    public Optional<CompletableFuture<Boolean>> publishTask(String name, Experiment experiment) throws TaskOperationException {
+    public CompletableFuture<Boolean> publishTask(String name, Experiment experiment) throws TaskOperationException {
         if (tasksOps.searchTask(name,experiment.getId()).isPresent())
             throw new TaskOperationException("Experiment is already published!");
-        return getPlatform(name).
-                map(platform1 -> platform1.publishTask(experiment)).
-                map(stringCompletableFuture -> stringCompletableFuture.handle((s, throwable) -> {
-                    if (s != null && throwable == null) {
-                        TaskRecord record = new TaskRecord();
-                        record.setExperiment(experiment.getId());
-                        record.setPlatformData(s);
-                        record.setStatus(TaskStatus.running);
-                        record.setCrowdPlatform(name);
-                        tasksOps.createTask(record);
-                    }
-                    return true;
-                }));
+
+        return getPlatform(name)
+                .map(platform1 -> platform1.publishTask(experiment))
+                .orElseThrow(() -> new IllegalArgumentException("Experiment not found!"))
+                .thenApply(s -> {
+                    TaskRecord record = new TaskRecord();
+                    record.setExperiment(experiment.getId());
+                    record.setPlatformData(s);
+                    record.setStatus(TaskStatus.running);
+                    record.setCrowdPlatform(name);
+                    return tasksOps.createTask(record) != null;
+                });
     }
 
     /**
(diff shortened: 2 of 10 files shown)


marcelhollerbach:
no review! problems detected!

from leander:
- make sure no duplicated tasks for experiment+platform
- getworkerid works now different

marcelhollerbach:
time for review.

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/crowdworking/PlatformManager.java
```diff
@@ -0,0 +1,217 @@
...
+
+    /**
+     * Publish the given experiment on the platform.
+     * The method will update the database with the new public task
+     *
+     * @param name The name of the platform
+     * @param experiment The experiment to publish
+     * @return None if the platform does not exist
+     */
+    public Optional<CompletableFuture<Boolean>> publishTask(String name, Experiment experiment) throws TaskOperationException {
```
LeanderK: this signature is not very elegant

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/crowdworking/PlatformManager.java
```diff
@@ -0,0 +1,217 @@
...
+        if (tasksOps.searchTask(name,experiment.getId()).isPresent())
+            throw new TaskOperationException("Experiment is already published!");
```
LeanderK: i already do that in the database, and my guarantees are better. But because i was smart, i never told anyone about it

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/crowdworking/PlatformManager.java
```diff
@@ -0,0 +1,217 @@
...
+                        TaskRecord record = new TaskRecord();
+                        record.setExperiment(experiment.getId());
+                        record.setPlatformData(s);
+                        record.setStatus(TaskStatus.running);
+                        record.setCrowdPlatform(name);
+                        tasksOps.createTask(record);
+                    }
+                    return true;
+                }));
+    }
```
LeanderK: i would suggest:

``` java
return getPlatform(name)
                .map(platform -> platform.publishTask(experiment))
                .orElse(CompletableFuture.supplyAsync(() -> {throw new IllegalArgumentException("platform not found");}))
                .thenApply(platformData -> {
                    TaskRecord record = new TaskRecord();
                    record.setExperiment(experiment.getId());
                    record.setPlatformData(platformData);
                    record.setStatus(TaskStatus.running);
                    record.setCrowdPlatform(name);
                    tasksOps.createTask(record);
                    return true;
                });
```

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/crowdworking/PlatformManager.java
```diff
@@ -0,0 +1,217 @@
...
+
+    /**
+     * Unpublish a given experiment from the given platform
+     *
+     * @param name The name of the platform
+     * @param experiment The experiment to unpublish
+     * @return None if the platform was not found, false if the unpublish failed and true if everything went fine
+     */
+    public Optional<CompletableFuture<Boolean>> unpublishTask(String name, Experiment experiment) throws TaskOperationException {
```
LeanderK: same as above

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/crowdworking/Platform.java
```diff
@@ -0,0 +1,58 @@
...
+     */
+    String getName();
+
+    /**
+     * Publish a passed experiment on the platform
+     * @param experiment the object which should be published
+     * @return This should return a unique string which is used to identify the published experiment later.
+     *         Will return None if the publishing failed
+     */
+    CompletableFuture<String> publishTask(Experiment experiment);
```
LeanderK: i would return an Future which completed exceptionally if the publishing failed, this makes the api easier to use

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/crowdworking/PlatformManager.java
```diff
@@ -0,0 +1,217 @@
...
+        TaskRecord record;
+
+        record = tasksOps.searchTask(name, experiment.getId()).
+                orElseThrow(() -> new TaskOperationException("Experiment is not published"));
+
+        return getPlatform(name).map(platform1 ->
+            platform1.unpublishTask(record.getPlatformData()).handle((b, throwable) -> {
```
LeanderK: if you only need the result and not the error (if there was one), i would suggest thenApply

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/crowdworking/PlatformManager.java
```diff
@@ -0,0 +1,217 @@
...
+        );
+    }
+
+    /**
+     * update the given experiment on the given platform
+     * @param name The name of the platform
+     * @param experiment The experiment to update
+     * @return None if the platform was not found, false if the update failed and true if everything went fine.
+     */
+    public Optional<CompletableFuture<Boolean>> updateTask(String name, Experiment experiment) throws TaskOperationException {
```
LeanderK: see the comments above

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/crowdworking/PlatformManager.java
```diff
@@ -0,0 +1,217 @@
...
+     *
+     * @param name The name of the platform
+     * @param experiment The experiment to unpublish
+     * @return None if the platform was not found, false if the unpublish failed and true if everything went fine
+     */
+    public Optional<CompletableFuture<Boolean>> unpublishTask(String name, Experiment experiment) throws TaskOperationException {
+        TaskRecord record;
+
+        record = tasksOps.searchTask(name, experiment.getId()).
+                orElseThrow(() -> new TaskOperationException("Experiment is not published"));
```
LeanderK: i would suggest multiple unpublishing results in no changes, this would simplify the api

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/crowdworking/PlatformManager.java
```diff
@@ -0,0 +1,217 @@
...
+        return getWorker(name).flatMap(worker -> worker.identifyWorker(params));
+    }
+
+    /**
+     * Get a worker record from the database which is associated with the given parameters
+     * @param name Name of the platform
+     * @param params Params which were passed by the platform to the workerservice
+     * @return A WorkerRecord if one is found
+     */
+    public Optional<WorkerRecord> getWorker(String name, Map<String, String[]> params) {
```
LeanderK: the hasEmail (or whatever you would like to name it) is missing

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/crowdworking/PlatformManager.java
```diff
@@ -0,0 +1,217 @@
...
+        return identifyWorker(name, params).map(uid ->
+            workerOps.getAllWorkers().stream().
+                    filter(workerRecord1 -> workerRecord1.getIdentification().equals(uid)).findFirst().
+                    orElseGet(() -> {
+                        //create a new entry in the database
+                        WorkerRecord record = new WorkerRecord(-1, uid, name, null);
+                        record = workerOps.createWorker(record);
+                        return record;
+                    })
```
LeanderK: it more readable if every stream command has its own line (also the dots are typically in the front):

``` java
workerOps.getAllWorkers().stream()
                   .filter(workerRecord1 -> workerRecord1.getIdentification().equals(did))
                   .findFirst()
                   .orElseGet(() -> {
                        //create a new entry in the database
                        WorkerRecord record = new WorkerRecord(-1, uid, name, null);
                        record = workerOps.createWorker(record);
                        return record;
                    })
```

i would also create a new workerOperations command to get one Worker getWorker(platform, identification)

``` java
workerOps.getWorker(platform, identification)
                  .orElseGet(() -> {
                        //create a new entry in the database
                        WorkerRecord record = new WorkerRecord(-1, uid, name, null);
                        record = workerOps.createWorker(record);
                        return record;
                    })
```

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/crowdworking/PlatformManager.java
```diff
@@ -0,0 +1,217 @@
...
+        );
+    }
+
+    public Optional<CompletableFuture<Boolean>> payWorker(String name, Worker worker, int amount) {
```
LeanderK: signature

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/crowdworking/PlatformManager.java
```diff
@@ -0,0 +1,217 @@
...
+    }
+
+    /**
+     * Will return the Worker interface which should be used to identify workers for the given platform
+     *
+     * @param name The name of the platform
+     * @return The interface used to identify a worker
+     */
+    public Optional<Worker> getWorker(String name) {
+        return getPlatform(name).map(platform -> platform.getWorker().orElse(fallbackWorker));
```
LeanderK: i would do a new line after every operation, but thats taste ;)

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/crowdworking/PlatformManager.java
```diff
@@ -0,0 +1,217 @@
...
+
+    /**
+     * Will return the payment service which should be used for the given platform
+     * If there is no Platform with the given name None is returned.
+     *
+     * @param name The name of the platform to use
+     * @return The interface used for payment
+     */
+    public Optional<Payment> getPlatformPayment(String name) {
+        return getPlatform(name).map(platform -> platform.getPayment().orElse(fallbackPayment));
```
LeanderK: above

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/crowdworking/Platform.java
```diff
@@ -0,0 +1,58 @@
...
+     * @param experiment The experiment with the new parameters
+     * @return The new id
+     */
+    CompletableFuture<String> updateTask(String id, Experiment experiment);
+
+    /**
+     * Returns if calibration questions can be asked on this platform
+     * @return true or false
+     */
+    Boolean isCalibsAllowd();
```
LeanderK: strange abbreviation

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/crowdworking/Worker.java
```diff
@@ -0,0 +1,16 @@
...
+/**
+ * Created by marcel on 19.01.16.
+ */
+public interface Worker {
+    /**
+     * Parse a worker id out of the params
+     * @param param The parameters which were sent by a platform
+     * @return The id of the worker if one can be found
+     */
+    Optional<String> identifyWorker(Map<String, String[]> param);
```
LeanderK: i would throw an exception instead of using an optional. something went terribly wrong if the Platform is unable to parse it's own input.

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/crowdworking/PlatformManager.java
marcelhollerbach: where do you check this ?

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/crowdworking/PlatformManager.java
marcelhollerbach: hasEmail is redundant for me, since i will get a email if one of the interfaces is not passed, so i dont care about that flag.

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/crowdworking/Platform.java
marcelhollerbach: tell it the one who patched the "entwurf ;)"

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/crowdworking/Worker.java
marcelhollerbach: yo.

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/crowdworking/PlatformManager.java
marcelhollerbach: as you like, not a big pain for me.

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/crowdworking/PlatformManager.java
marcelhollerbach: ok

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/crowdworking/Platform.java
LeanderK: i would still not use it in the actual code

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/crowdworking/PlatformManager.java
LeanderK: doIfNotRunning

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/crowdworking/PlatformManager.java
LeanderK: no its not, if the platform needs an email and hasEmail is false no worker must be created

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/crowdworking/PlatformManager.java
marcelhollerbach: well this would mean i need to create the task -> check if not failed -> publish on crowdplatform -> remove if failed -> insert data if not -> update which is like 100 times harder than doing that. maybe make ifNotRunnign public as boolean so i can use it ?

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/crowdworking/PlatformManager.java
marcelhollerbach: do you have some writen spec on what this api should do ?

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/crowdworking/PlatformManager.java
LeanderK: ah ok, yeah i forgot that. it would first create the task and then publish it on the crowdPlatform (and if failed remove it) because of concurrency

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/crowdworking/PlatformManager.java
LeanderK: getWorker:
used by the WorkerService to get the internal representation of the worker it communicates with. It first searches for the worker in the database and if not found may return a new one (see hasEmail).
@param String platform: the crowdworking platform the worker is working in
@param Map<String, String[]> data: passed platform-specific data from the worker, may be used to identify the worker by the platform
@param boolean hasEmail: if the crowdplatform needs an email, but the workerservice does not have one right now the method should only return a worker if it is already existing (and has an email). if not the parameter is always false and the platform should create a new one

marcelhollerbach: this sounds like a poor api, why only create a worker under this special cases ?

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/crowdworking/PlatformManager.java
marcelhollerbach: where is a concurrency problem ?

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/crowdworking/PlatformManager.java
LeanderK: i was thinking about it. I don't think its that bad, just look at java.util.map :P
we can split it, but the implementation into getExistingWorker and getWorker, but the implementation of the method above is pretty trivial (just look at the concept, just guard the create worker block with `(hasNativePayment || hasEmail)`)

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/crowdworking/PlatformManager.java
LeanderK: if you publish a task twice then on the if clause would fail for both requests and therefore not throw an exception. I don't what happens then with the whole system. Therefore you don`t check mutable things like that if you change them afterwards

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/crowdworking/PlatformManager.java
marcelhollerbach: what i am wondering about is: A crowdplatform should identify a worker. The logic under which cases a user should be created or just read is not the job of the platform management. Thats a implementation case of a special rest call or some implementation in the workerservice. 
tl;dr this logic does not belong in this place here.

LeanderK: i mean you can move the logic around, but i think i may belong here. i mean the platform-package is concerned with platform-stuff and this is platform stuff, independent of the db-call.  But we can move the db logic into the rest-call area and create a getPlatformData method that takes the platform and returns the workerRecord and another that returns the string. (i would really like to not be so dependent of the exact matching of the string, because thats more the limitations of the current database and we may get rid of it in the future, giving us far greater flexibility)

marcelhollerbach: the platform package is concerned with platform stuff, true. the only platform stuff here is the identification. The reason why i think this does not belong here: lets say the ws is changing, also the rest call. you would have to change the platform package, which is somehow wrong, nothing in the platform has changed, only the rest call and the ws.

LeanderK: i mean my idea above about getWorker and createPlatformData (maybe better than get). They are not rest calls.

marcelhollerbach: sure, they are not direct rest calls, but the logic of the call is very depending on the way how the ws works. And has no basic usage like publishTask or something like that. The way how the method works is very depending on the ws logic. Thats why I would like to bind this logic to the rest call, will add a patch if this branch is landed.

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/crowdworking/Platform.java
kelunik: Misses an `e` in `Allowed`, too.

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/crowdworking/PlatformManager.java
LeanderK: getWorker and createPlatformData data is basic logic, you can't divide it up to a more primitive call. The rest call combines these primitives to whatever the deepest desires of the ws fulfils. Also its purely platform-dependent, i can`t do anything in the rest-package

marcelhollerbach: well for the platform the basic logic is to identify the worker. A call like identifyWorker. The logic in which cases a new record is created or not can be done in the rest call.

LeanderK: getWorker gets a worker if he is existing, createPlatformData would just generate what to save with the data. only queries, no inserts if thats so important to you. I really don`t want to rely on string equality in the api that much (therefore getWorker), it will limit future flexibility heavily.

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/crowdworking/Platform.java
LeanderK: thank you captain obvious

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/crowdworking/PlatformManager.java
marcelhollerbach: I added the code, and in the end in the two functions nothing else than db operations are done, which are not really something which needs to be done in the manager. Those db calls could have also been called in the rest handler.

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/crowdworking/PlatformManager.java
```diff
@@ -0,0 +1,240 @@
...
+            throw new TaskOperationException("Task could not be created");
+
+        return getPlatform(name)
+                .map(platform1 -> platform1.publishTask(experiment))
+                .orElseThrow(() -> new IllegalArgumentException("Platform not found!"))
+                .handle((s1, throwable) -> {
+                    if (s1 != null) {
+                        result.setPlatformData(s1);
+                    } else {
+                        result.setStatus(TaskStatus.stopped);
```
LeanderK, line 137: you have to call method in the operations package. This will change nothing in the database and i don`t guarantee that .store will work

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/crowdworking/PlatformManager.java
```diff
@@ -0,0 +1,240 @@
...
+     * Unpublish a given experiment from the given platform
+     *
+     * @param name The name of the platform
+     * @param experiment The experiment to unpublish
+     * @return None if the platform was not found, false if the unpublish failed and true if everything went fine
+     */
+    public CompletableFuture<Boolean> unpublishTask(String name, Experiment experiment) throws TaskOperationException {
+        TaskRecord record;
+
+        record = tasksOps.searchTask(name, experiment.getId()).orElse(null);
```
LeanderK: i would first set it to unpublished, so no other thread can unpublish it at the same time

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/crowdworking/PlatformManager.java
```diff
@@ -0,0 +1,240 @@
...
+        TaskRecord record;
+
+        record = tasksOps.searchTask(name, experiment.getId()).
+                orElseThrow(() -> new TaskOperationException("Experiment is not published"));
+
+        return getPlatform(name)
+                .map(platform -> platform.updateTask(record.getPlatformData(), experiment))
+                .orElseThrow(() -> new IllegalArgumentException("Platform not found"))
+                .thenApply(s -> {
+                    record.setPlatformData(s);
```
LeanderK, line 185: same as above

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/crowdworking/PlatformManager.java
```diff
@@ -0,0 +1,240 @@
...
+    }
+
+    /**
+     * Get a worker if he exists
+     * @param name Name of the platform
+     * @param params Params passed by the platform
+     * @return A worker if one is found
+     * @throws UnknownWorkerException if the platform does not identify a worker
+     */
+    public Optional<WorkerRecord> getWorker(String name, Map<String ,String[]> params) throws UnknownWorkerException {
```
LeanderK: i don`t get the signature, why not return empty if not found?

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/crowdworking/PlatformManager.java
marcelhollerbach, line 137: wad ?

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/crowdworking/PlatformManager.java
marcelhollerbach, line 185: updateTask does _not_ update stuff ?

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/crowdworking/Worker.java
```diff
@@ -0,0 +1,15 @@
+package edu.kit.ipd.crowdcontrol.objectservice.crowdworking;
+
+import java.util.Map;
+
+/**
+ * Created by marcel on 19.01.16.
+ */
+public interface Worker {
```
LeanderK: i don`t think the name is optimal, i would name the interface IdentifyWorker and also mark it with @FunctionalInterface.

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/crowdworking/PlatformManager.java
marcelhollerbach: it returns none if not found, and the exception if the platfrom can not identify the user

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/crowdworking/PlatformManager.java
LeanderK, line 137: this class only holds the results of the query. you can manipulate the data in TaskRecord, but that won't change the database.

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/crowdworking/PlatformManager.java
LeanderK, line 185: no, i mean record stuff not changing the database

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/crowdworking/PlatformManager.java
marcelhollerbach, line 137: i recommend reading the next lines of code

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/crowdworking/PlatformManager.java
LeanderK: that seems a bit clunky, why not return empty if no worker was found? and if you really want to use the exceptions, it should be the other way around (UnkownPlatformException and empty worker if worker not found)

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/crowdworking/PlatformManager.java
LeanderK: i have read it, but you whole method never changes anything in the database

LeanderK, line 137: i have read it, but your method publishes the task, but never updates its content. The platformData will be lost when the method returns

marcelhollerbach, line 137: i have absolutly no idea what you are talking about. if                     

```
record.setPlatformData(s);
return tasksOps.updateTask(record);
```

Do you want to tell me this is not updating the database ?

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/crowdworking/PlatformManager.java
marcelhollerbach: empty is returned if not worker is found in the database, and a exception if the platform failes to identify the user (which means the whole request is faulty)

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/crowdworking/PlatformManager.java
LeanderK, line 137: yeah, the record instances are the results of queries. They are mutable but changing their data does not change the data in the database. They only represent the results. There is a method .store, but i don't guarantee that it's working, so don't use it.

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/crowdworking/PlatformManager.java
LeanderK: ah ok, then i would rename the Exception

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/crowdworking/PlatformManager.java
marcelhollerbach, line 137: ok - you tell me updateTask does not update the databse and .store maybe but maybe not. So how do I use this magic bit called db ? :>

LeanderK, line 137: just create a method in the database package, like updateTask(TaskRecord record)

marcelhollerbach, line 137: WAD ? like the one which exist and is called 2 lines later ? WTF

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/crowdworking/PlatformManager.java
marcelhollerbach: no, just read the comment in which case the exception is thrown then there is no need to rename the exception

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/crowdworking/PlatformManager.java
LeanderK, line 137: ok i am now officially stupid and unable to review code. yeah, this is the method :D I should probably take a break...

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/crowdworking/PlatformManager.java
LeanderK, line 185: ok, also my fault

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/crowdworking/PlatformManager.java
marcelhollerbach, line 137: go and get some air!!! :+1:

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/crowdworking/PlatformManager.java
LeanderK: no, the name of the execution must describe its purpose. often these exceptions get propagated through the system so you have to rely on good names to understand where the exception is coming from and what it is doing.

LeanderK:
i think it's good to go, but i don`t trust my decisions right now. I will go for a run and then sit down and review it seriously, so you hear from me in 45 minutes. I think the residual alcohol from yesterday is not helping either.

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/crowdworking/PlatformManager.java
kelunik: I think `UnknownWorkerException` is fine, `WorkerIdentificationFailedException` might be an alternative name.

LeanderK: i think the new name is better

kelunik: What's the new / old name?

PR #27, opened 2016-01-23
database: Consistency is all I ask! Immortality is all I seek. 

LeanderK:
- changed some names of methods in `NotificationRestOperations`, `NotificationOperations` and one in `TasksOperations`

some housekeeping-stuff in the API.

### DIFF
--- a/src/main/java/edu/kit/ipd/crowdcontrol/objectservice/database/operations/TasksOperations.java
+++ b/src/main/java/edu/kit/ipd/crowdcontrol/objectservice/database/operations/TasksOperations.java
@@ -49,7 +49,7 @@ public class TasksOperations extends AbstractOperations {
      * @param experimentId the primary key of the experiment
      * @return the found task or empty if not found
      */
-    public Optional<TaskRecord> searchTask(String platform, int experimentId) {
+    public Optional<TaskRecord> findTask(String platform, int experimentId) {
         return create.selectFrom(Tables.TASK)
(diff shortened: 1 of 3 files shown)


src/main/java/edu/kit/ipd/crowdcontrol/objectservice/database/operations/TasksOperations.java
```diff
@@ -49,7 +49,7 @@ public boolean updateTask(TaskRecord taskRecord) throws IllegalArgumentException
      * @param experimentId the primary key of the experiment
      * @return the found task or empty if not found
      */
-    public Optional<TaskRecord> searchTask(String platform, int experimentId) {
+    public Optional<TaskRecord> findTask(String platform, int experimentId) {
```
marcelhollerbach: there was somewhere a cool blog entry about using search vs. find. The message of the entry was that you should only use find if you can assert that something will be found. according to that i would use search here, since the search can fail.

LeanderK: i have read something about find being one and search being multiple items. I don't think it's that clearly defined since most of the apis i remember use find (including the all ubiquitous java regex-api). But i think i was mislead by the original name. Consistent would be get ;) wait for a new commit. It`s a bit of a mess, because get is the standard java, fetch used in the database-drivers to execute queries and search/find would be possible too.

marcelhollerbach:
Ok

PR #30, opened 2016-01-25
database: updated model, fixed DatabaseManager init-code

LeanderK:
- the model now includes the feedback-column in rating
- updated the db.sql script
- (hopefully) fixed DatabaseManager.initDatabase()


kelunik:
`]; You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'USE`crowdcontrol`;`

Seems like multi-statements are not supported?

LeanderK:
what is your jdbc-url? should be like: `jdbc:mysql://localhost:3306/crowdcontrol?allowMultiQueries=true`

LeanderK:
hmm...i might ne able to make this unnecessary

kelunik:
With `allowMultiQueries=true`:

```
Error executing: CREATE DEFINER = CURRENT_USER TRIGGER `crowdcontrol`.`Platform_BEFORE_INSERT` BEFORE INSERT ON `Platform` FOR EACH ROW
BEGIN
    SET NEW.id_platform = LOWER(NEW.id_platform);
    IF (NEW.id_platform REGEXP '^[a-z0-9_]+$') = 0 THEN
        SIGNAL SQLSTATE '45000' SET MESSAGE_TEXT = "id_platform does not match regex";
    END IF;
END
.  Cause: com.mysql.jdbc.exceptions.jdbc4.MySQLSyntaxErrorException: This version of MySQL doesn't yet support 'multiple triggers with the same action time and event for one table'
```

LeanderK:
hmm strange, there should not be multiple triggers with the same action time and event

kelunik:
I think the previous schema isn't properly deleted.

LeanderK:
hmm there is not much ambiguity left with `DROP DATABASE crowdcontrol;`

LeanderK:
but it looks like it is
EDIT: i think i know why, the small subtleties of JDBC are a bit annoying

LeanderK:
@kelunik can you test it now without multiple queries activated? The drop should work now and locally it initialises the database correctly.

kelunik:
My database user `crowdcontrol` doesn't have global rights, so I need to execute `CREATE SCHEMA IF NOT EXISTS crowdcontrol DEFAULT CHARACTER SET utf8mb4 ;` manually. After doing that, it seems to work.

marcelhollerbach:
looks sane to me

PR #31, opened 2016-01-26
API: Worker

kelunik:
Not ready for merge yet!


src/main/java/edu/kit/ipd/crowdcontrol/objectservice/Main.java
```diff
@@ -65,11 +67,15 @@ private static void boot(Connection connection) {
         TemplateOperations templateOperations = new TemplateOperations(context);
         NotificationRestOperations notificationRestOperations = new NotificationRestOperations(context);
         PlatformOperations platformOperations = new PlatformOperations(context);
+        WorkerOperations workerOperations = new WorkerOperations(context);
+
+        // TODO Instantiate PlatformManager
```
kelunik, line 72: @marcelhollerbach just a default instance will do for now?

kelunik:
Completely bogus rebase, will start from scratch…

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/Main.java
marcelhollerbach, line 72: yep, I think this is the best place for it, since the object can be handed to every part where it is needed.

PR #32, opened 2016-01-26
API: Worker

kelunik:
Not ready for merge yet!

### DIFF
--- a/src/main/java/edu/kit/ipd/crowdcontrol/objectservice/rest/resources/WorkerResource.java
+++ b/src/main/java/edu/kit/ipd/crowdcontrol/objectservice/rest/resources/WorkerResource.java
@@ -34,16 +34,13 @@ public class WorkerResource {
      * @return Worker if found.
      */
     public Worker identify(Request request, Response response) {
-        String identity;
-
         try {
-            identity = manager.identifyWorker(request.params("platform"), request.queryMap().toMap());
+            return manager.getWorker(request.params("platform"), request.queryMap().toMap())
+                    .map(WorkerOperations::toProto)
+                    .orElseThrow(() -> new NotFoundException("Resource not found."));
         } catch (UnidentifiedWorkerException e) {
             throw new BadRequestException("Unidentified worker!");
         }
-
-        return operations.identifyWorker(request.params("platform"), identity)
-                .orElseThrow(() -> new NotFoundException("Resource not found."));
     }
 
     /**
--- a/src/main/java/edu/kit/ipd/crowdcontrol/objectservice/rest/resources/WorkerResource.java
+++ b/src/main/java/edu/kit/ipd/crowdcontrol/objectservice/rest/resources/WorkerResource.java
@@ -1,0 +1,115 @@
+package edu.kit.ipd.crowdcontrol.objectservice.rest.resources;
+
+import edu.kit.ipd.crowdcontrol.objectservice.crowdworking.PlatformManager;
+import edu.kit.ipd.crowdcontrol.objectservice.crowdworking.UnidentifiedWorkerException;
+import edu.kit.ipd.crowdcontrol.objectservice.database.operations.WorkerOperations;
+import edu.kit.ipd.crowdcontrol.objectservice.proto.Worker;
+import edu.kit.ipd.crowdcontrol.objectservice.proto.WorkerList;
+import edu.kit.ipd.crowdcontrol.objectservice.rest.Paginated;
+import edu.kit.ipd.crowdcontrol.objectservice.rest.exceptions.BadRequestException;
+import edu.kit.ipd.crowdcontrol.objectservice.rest.exceptions.NotFoundException;
+import spark.Request;
+import spark.Response;
+
+import static edu.kit.ipd.crowdcontrol.objectservice.rest.RequestUtil.*;
+
+/**
+ * Handles requests to worker resources.
+ *
+ * @author Niklas Keller
+ */
+public class WorkerResource {
+    private WorkerOperations operations;
+    private PlatformManager manager;
+
+    public WorkerResource(WorkerOperations operations, PlatformManager manager) {
+        this.operations = operations;
+        this.manager = manager;
+    }
+
+    /**
+     * @param request  request provided by Spark.
+     * @param response response provided by Spark.
+     *
+     * @return Worker if found.
+     */
+    public Worker identify(Request request, Response response) {
+        String identity;
+
+        try {
+            identity = manager.identifyWorker(request.params("platform"), request.queryMap().toMap());
+        } catch (UnidentifiedWorkerException e) {
+            throw new BadRequestException("Unidentified worker!");
+        }
+
+        return operations.identifyWorker(request.params("platform"), identity)
+                .orElseThrow(() -> new NotFoundException("Resource not found."));
+    }
+
+    /**
+     * @param request  request provided by Spark.
+     * @param response response provided by Spark.
+     *
+     * @return A list of all workers.
+     */
+    public Paginated<Integer> all(Request request, Response response) {
+        int from = getQueryInt(request, "from", 0);
+        boolean asc = getQueryBool(request, "asc", true);
+
+        return operations.getWorkerList(from, asc, 20)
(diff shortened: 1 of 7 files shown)


src/main/java/edu/kit/ipd/crowdcontrol/objectservice/rest/resources/WorkerResource.java
```diff
@@ -0,0 +1,115 @@
...
+    /**
+     * @param request  request provided by Spark.
+     * @param response response provided by Spark.
+     *
+     * @return {@code null}.
+     */
+    public Worker delete(Request request, Response response) {
+        boolean existed = false;
+
+        // TODO operations.anonymizeWorker(getParamInt(request, "id")); should accept int and return boolean
```
kelunik, line 104: @LeanderK According to description, it already deletes the worker. It should return `false` if the worker didn't exist. Additionally, we need another identify here then?

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/rest/resources/WorkerResource.java
```diff
@@ -0,0 +1,115 @@
...
+     * @param request  request provided by Spark.
+     * @param response response provided by Spark.
+     *
+     * @return Worker if found.
+     */
+    public Worker identify(Request request, Response response) {
+        String identity;
+
+        try {
+            identity = manager.identifyWorker(request.params("platform"), request.queryMap().toMap());
```
marcelhollerbach: Can you use  the manager function which returns a worker record ?

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/rest/resources/WorkerResource.java
marcelhollerbach, line 104: I think its better to check if the worker can be found before. Delete should only return false if the operation failed.

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/rest/resources/WorkerResource.java
kelunik: Then I either need to move `toProto` or make it `public static`.

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/rest/resources/WorkerResource.java
kelunik, line 104: If the operation failed, it should throw an exception.

marcelhollerbach: I think public static is fine, since the methods can useful in other pieces of code

marcelhollerbach: Wrong comment ....

kelunik: But we should move it out of operations then. Into `WorkerResource` or another class?

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/rest/resources/WorkerResource.java
marcelhollerbach: I think public static is fine, since the methods can useful in other pieces of code

kelunik: But we should move it out of operations then. Into `WorkerResource` or another class?

marcelhollerbach: If you see the operations as until class for experiments etc. Then its fine, but we could wrote such methods in something like experimentUtil where just plain converts are done

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/rest/resources/WorkerResource.java
marcelhollerbach, line 104: I don't think so. Its basically a query and the result should indicate if the deletion was OK or not - You could even suspect that deletion of a unknown worker would return true since the deletion was "successfully" the entry with the is is gone.

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/rest/resources/WorkerResource.java
kelunik: Made them `public static` in all operation classes.

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/rest/resources/WorkerResource.java
kelunik, line 104: You didn't delete anything if there was nothing. ;-)

marcelhollerbach, line 104: But The request is full filled, the entry with is x is not there (anymore)

marcelhollerbach, line 104: Is=id 
What I want to say: deleting a entry can have more than one correct behavior, depending on your use case. And with checking if the resource exists you can make sure that the request is even valid. So no need to be scared that the db could return something else than expected.

kelunik, line 104: I think it's usual to return a `404` if the resource has already been deleted. 
http://stackoverflow.com/a/24713946/2373138

marcelhollerbach, line 104: You just don't get what I am talking about. I am talkimg about the database could return true from a delete request since the request itself is fulfilled. 
In the end: just check if the resource exists.

kelunik, line 104: We have the data while deleting, because we can check the affected rows. Checking it previously creates a possible race condition which is simple to avoid if we add a return value here.

marcelhollerbach, line 104: where is a race confition if you just check if the worker exists with the given id ?

kelunik, line 104: `Request 1: DELETE /workers/1`
`Request 2: DELETE /workers/1`

`1: exist? yes` → `2: exists? yes` → `1: delete!` → `2: delete!`

instead of:

`1: delete? yes` → `2: delete? no` → `1: deleted!` → `2: not found!`

marcelhollerbach, line 104: mhmm ok - but there is still a problem, since a failed anonymizeWorker does not indicate a resource which is not there.

kelunik, line 104: Which failure are we talking about?

marcelhollerbach, line 104: false returned by anonymizeWorker

kelunik, line 104: Sorry, asked about the case. Why should anything other return `false`?

marcelhollerbach, line 104: connection problems, triggers to delete other tables which failed, internal database computation problems... maybe just return a int which indicates the sum of changes or -1 for the error case?

kelunik, line 104: All these should be exceptions and not return values.

PR #40, opened 2016-01-27
Database

LeanderK:
- updated db.sql (no changes on the model)
- this is my first pull request without a new model-version!
- added needed queries for the payment package
- fixed some protobuf-related bugs.
- i should mention that there is no new model


src/main/resources/db.sql
```diff
@@ -92,8 +92,8 @@ CREATE TABLE IF NOT EXISTS `crowdcontrol`.`Experiment` (
   CONSTRAINT `usedTemplate`
     FOREIGN KEY (`template`)
     REFERENCES `crowdcontrol`.`Template` (`id_template`)
-    ON DELETE NO ACTION
-    ON UPDATE NO ACTION,
+    ON DELETE RESTRICT
+    ON UPDATE CASCADE,
```
kelunik, line 96: Do we really want to restrict deletion?

LeanderK, line 96: yes, thats coherent with the rest of the database. Our foreign-keys are valid or null. When i introduce an exception here it gets messy very quickly.

kelunik, line 96: Sure, if, then remove the constraint all together.

LeanderK, line 96: but then we have no guarantees. This is mainly there to prevent bugs, where one forgets to remove the key before deleting the template.

kelunik, line 96: I'm talking about removing the `template_id` from experiments.

LeanderK, line 96: we already talked about this extensively ;) No, it stays

PR #45, opened 2016-01-28
Fix and improve error messages and docs

kelunik:


marcelhollerbach:
Is this PR reviewd by yourself ? :-D I cannot see any tag from @LeanderK or someone else ...

kelunik:
Except for new `NotFoundException` constructor, it's just formatting `@param` into the right format. I don't think that needs review. Talked enough with @LeanderK about parameter formatting.

marcelhollerbach:
You wrote 2 comments above that this PR was reviewed. Which is obviously not true.

And also the patch also includes the change from notification to template in a route, and errormessage changes. Its okay to have them in one PR, what I want to say is to split them up into several commits.

And B.T.W it would be cool if you could just attach the persons which are working with this exceptions to the commit (or PR) so they can see the new api.

kelunik:
I meant "reviewed" as that others will see the change, because you said the commits should have been split.

marcelhollerbach:
dont understand why a closed PR should be reviewed by anyone.

PR #49, opened 2016-01-29
Mail2

TheRittler:
mail package improved

### DIFF
--- a/src/integration-test/java/edu/kit/ipd/crowdcontrol/objectservice/mail/MailHandlerTest.java
+++ b/src/integration-test/java/edu/kit/ipd/crowdcontrol/objectservice/mail/MailHandlerTest.java
@@ -3,6 +3,9 @@ package edu.kit.ipd.crowdcontrol.objectservice.mail;
 import org.junit.Test;
 
 import javax.mail.Message;
+import java.io.BufferedInputStream;
+import java.io.FileInputStream;
+import java.util.Properties;
 import java.util.UUID;
 
 import static org.junit.Assert.assertTrue;
@@ -21,7 +24,12 @@ public abstract class MailHandlerTest {
         String uuid = UUID.randomUUID().toString();
         String subject = "[test] " + uuid;
 
-        handler.sendMail("pseipd@trash-mail.com", subject, uuid);
+        Properties properties = new Properties();
+        BufferedInputStream stream = new BufferedInputStream(new FileInputStream("src/integration-test/resources/mailHandlerTestProps.properties"));
+        properties.load(stream);
+        stream.close();
+
+        handler.sendMail(properties.getProperty("receiver"), subject, uuid);
         try {
             Thread.sleep(3000);
         } catch (InterruptedException e) {}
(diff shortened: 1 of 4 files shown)


src/integration-test/java/edu/kit/ipd/crowdcontrol/objectservice/mail/MailHandlerTest.java
```diff
@@ -21,11 +21,11 @@ public void test() throws Exception {
         String uuid = UUID.randomUUID().toString();
         String subject = "[test] " + uuid;
 
-        handler.sendMail(this.mail, subject, uuid);
+        handler.sendMail("pseipd@trash-mail.com", subject, uuid);
```
LeanderK: should this be hardcoded?

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/mail/MailHandler.java
```diff
@@ -39,22 +40,39 @@ public MailHandler(Properties props, Authenticator auth) throws MessagingExcepti
         connectToStore();
```
LeanderK: you open the resource, but i don't see where you close it.

src/integration-test/java/edu/kit/ipd/crowdcontrol/objectservice/mail/MailHandlerTest.java
kelunik: Nope, should be in a config file.

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/mail/MailHandler.java
TheRittler: If I close the store, you cant use the messages anymore.

LeanderK: yeah, but you can reopen them when you need them, i wouldn't leave it open 24/7

TheRittler: the method returns a Message[] Array, you cant even read the messages without connected store. So the user of the messages has to open the store via message[i].getFolder(),getStore().open() before he can read that, and he has to close it after use. I can implement that this way and write it to the javadocs, but its a bit uncomfortable

kelunik: I'd just use a resource try / catch like

``` java
try (connection = handler.connect()) {
    // fetch mails and read them here
} catch (IOException e) {
    // oops, something went wrong
}
```

Resource try / catch automatically closes the resource if it's autoclosable and was open.

TheRittler: it is not autoclosable ;)

LeanderK: then you have to close it manually! even if there are now exceptions, don't forget the finally block.

TheRittler: I solved the problem, but not the way you wanted, because the other way I got lots of spaghetti code. Now the store has the same lifetime as a message or the folder of the message.

LeanderK: it's generally bad practice to let the gc close the resources for you.

TheRittler:
i cant implement closable, because then i have to implement close(), but the close method has parameters.

PR #51, opened 2016-01-31
update Event handling

marcelhollerbach:
this commit introduces the event class which is emitted by every event
emit. With the Event class you have access to the EventLog which is
helpfull for adding log messages for a subscriber.

### DIFF
--- a/src/main/java/edu/kit/ipd/crowdcontrol/objectservice/event/EventLog.java
+++ b/src/main/java/edu/kit/ipd/crowdcontrol/objectservice/event/EventLog.java
@@ -2,6 +2,7 @@ package edu.kit.ipd.crowdcontrol.objectservice.event;
 
 import java.util.ArrayList;
 import java.util.List;
+import java.util.concurrent.ConcurrentLinkedQueue;
 
 /**
  * Can be used to create a stack of messages from a event subscriber.
@@ -10,13 +11,13 @@ import java.util.List;
  */
 public class EventLog {
     private String name;
-    private List<LogEntry> logs;
+    private ConcurrentLinkedQueue<LogEntry> logs;
 
     /**
      * Create a new Log
      */
     public EventLog() {
-        logs = new ArrayList<>();
+        logs = new ConcurrentLinkedQueue<>();
         name = "";
     }
 
--- a/src/main/java/edu/kit/ipd/crowdcontrol/objectservice/event/EventLog.java
+++ b/src/main/java/edu/kit/ipd/crowdcontrol/objectservice/event/EventLog.java
@@ -1,6 +1,6 @@
 package edu.kit.ipd.crowdcontrol.objectservice.event;
 
-import java.util.LinkedList;
+import java.util.ArrayList;
 import java.util.List;
 
 /**
@@ -16,7 +16,7 @@ public class EventLog {
      * Create a new Log
      */
     public EventLog() {
-        logs = new LinkedList<>();
+        logs = new ArrayList<>();
         name = "";
     }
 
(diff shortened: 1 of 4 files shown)


src/main/java/edu/kit/ipd/crowdcontrol/objectservice/event/EventLog.java
```diff
@@ -0,0 +1,79 @@
...
+        this.name = name;
+    }
+
+    /**
+     * Add a new message to this stack
+     * @param shortmessage a short explanation on what this message is about. No linebreaks allowed.
+     * @param longmessage detailed message what happends exactly. Linebreaks allowed.
+     */
+    public synchronized void addEntry(String shortmessage, String longmessage) {
+        logs.add(new LogEntry(shortmessage, longmessage));
```
LeanderK: you can ignore my comment, but do we don`t really need a list, since you are never using an index. For these concurrent things there is a ConcurrentLinkedQueue, which provides nonblocking FIFO operations

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/event/EventObservable.java
```diff
@@ -25,16 +27,25 @@ public EventObservable() {
...
      * Will emit the given object as event to the subscribed subscribers of the observable.
      * @param t The object to pass to the events
+     * @return The messages returned by the emits
      */
-    public void emit(T t) {
-        subs.stream().forEach(subscriber -> subscriber.onNext(t));
+    public List<EventLog> emit(T t) {
+        return subs.stream()
+                .map(subscriber -> eventEmit(subscriber, t))
+                .collect(Collectors.toList());
```
LeanderK, line 35: this is sequential, i don't think we want this to be sequential. you can use parStream() to achieve concurrency

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/event/EventObservable.java
```diff
@@ -4,14 +4,16 @@
...
 
 /**
  *
  * Created by marcel on 02.01.16.
  */
 public class EventObservable<T> {
-    private final Observable<T> observable;
-    private final LinkedList<Subscriber<? super T>> subs;
+    private final Observable<Event<T>> observable;
+    private final LinkedList<Subscriber<? super Event<T>>> subs;
```
LeanderK, line 16: you don't want to use LinkedLists

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/event/EventObservable.java
marcelhollerbach, line 35: this pr is about adding a eventlog not about parallel or not...

LeanderK, line 35: but you added the code and when i give the review its in the master

marcelhollerbach, line 35: yeah and then there will be the next pr which introduces parallel emittion

LeanderK: turning stream() into parallelStream() would fix that, i don't think you need a pr for this

LeanderK, line 35: hmm ok, I'll remember this statement ;)

marcelhollerbach, line 35: ok :)

PR #52, opened 2016-01-31
platformmanager: correctly handle fail cases

marcelhollerbach:
if a exception is thrown we need to rethrow it. If the platform does not
  give any usable key we also need to throw up here

### DIFF
--- a/src/main/java/edu/kit/ipd/crowdcontrol/objectservice/crowdworking/PlatformManager.java
+++ b/src/main/java/edu/kit/ipd/crowdcontrol/objectservice/crowdworking/PlatformManager.java
@@ -131,6 +131,8 @@ public class PlatformManager {
      * Publish the given experiment on the platform.
      * The method will update the database with the new public task
      *
+     * If a exception is thrown NO TaskRecord is created, this means the Experiment will still be known as "unpublished"
+     *
      * @param name The name of the platform
      * @param experiment The experiment to publish
      * @return None if the platform does not exist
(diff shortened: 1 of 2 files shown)


marcelhollerbach:
In theory this has nothing to do with the later publish unpublish stuff - just to know that I dont break any convention :-D

marcelhollerbach:
@phimonk handling better now ?

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/crowdworking/PlatformManager.java
```diff
@@ -150,13 +152,22 @@ public Payment getPlatformPayment(String name) {
...
+                    if (s1 != null && throwable == null) {
                         result.setPlatformData(s1);
-                    } else {
-                        result.setStatus(TaskStatus.stopped);
+                        if (!tasksOps.updateTask(result)) {
+                            throw new IllegalStateException("Updating record for published task failed");
+                        }
+                    }
+                    //if there is no useful key throw!
+                    if (s1 == null || s1.isEmpty()) {
```
simonkorz: I just figured out that if an exception occured, s1 is indeed always null, no matter if it was assigned a value or not. So you were right! That means the third "if block" will never be executed. So it might be a good idea to combine the last two "if blocks". I think it is important to pass or handle the errors that occur on the platform connection level.

marcelhollerbach: Mhmm then just the s1 != null is not needed, s1 could be empty which would result in a for us error case.

simonkorz: Agree

LeanderK:
push @marcelhollerbach

marcelhollerbach:
pushed.

PR #53, opened 2016-01-31
Notification

simonkorz:
Emails might not be toooo beatiful, but I think it's ready


LeanderK:
i can review it

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/notification/NotificationController.java
```diff
@@ -0,0 +1,106 @@
...
+
+/**
+ * The NotificationController holds all notifications and schedules them.
+ *
+ * @author Simon Korz
+ * @version 1.0
+ */
+public class NotificationController {
+    private final ScheduledExecutorService scheduler =
+            Executors.newScheduledThreadPool(1);
```
LeanderK: this line-wrapping is bit unnecessary

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/notification/NotificationController.java
```diff
@@ -0,0 +1,106 @@
...
+     * Deletes a notification
+     *
+     * @param notification the notification to delete
+     */
+    public void deleteNotification(Notification notification) {
+        deleteNotification(notification.getID());
+    }
+
+    private void deleteNotification(int id) {
+        ScheduledFuture<?> notificationHandle = handleMap.get(id);
```
LeanderK: you get an NullpointerException if the id is not existing in the map, but i don't know if that can happen

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/notification/NotificationController.java
```diff
@@ -0,0 +1,106 @@
...
+                scheduler.scheduleAtFixedRate(notification, 0, notification.getCheckPeriod(), TimeUnit.SECONDS);
+        handleMap.put(notification.getID(), notificationHandle);
+    }
+
+    /**
+     * Deletes a notificationProto
+     *
+     * @param notificationProto the notification to delete
+     */
+    public void deleteNotification(edu.kit.ipd.crowdcontrol.objectservice.proto.Notification notificationProto) {
```
LeanderK, line 78: not used anywhere...do you need it?

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/notification/Notification.java
```diff
@@ -0,0 +1,131 @@
...
+import java.time.Instant;
+import java.time.temporal.ChronoUnit;
+
+/**
+ * The Notification class represents a complete Notification.
+ *
+ * @author Simon Korz
+ * @version 1.0
+ */
+public class Notification implements Runnable {
```
LeanderK, line 12: its a design choice, but you could just reference the NotificationRecord

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/notification/NotificationController.java
```diff
@@ -0,0 +1,106 @@
...
+     *
+     * @param notification the notification to delete
+     */
+    public void deleteNotification(Notification notification) {
+        deleteNotification(notification.getID());
+    }
+
+    private void deleteNotification(int id) {
+        ScheduledFuture<?> notificationHandle = handleMap.get(id);
+        notificationHandle.cancel(true);
```
LeanderK: i would set this to false...it doesn't hurt to let this run for one more time and you don't have to be extra careful closing you resources

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/notification/NotificationPolicy.java
```diff
@@ -0,0 +1,34 @@
+package edu.kit.ipd.crowdcontrol.objectservice.notification;
+
+/**
+ * NotificationPolicy encapsulates strategies to check and send notifications.
+ *
+ * @author Simon Korz
+ * @version 1.0
+ */
+public abstract class NotificationPolicy<T> {
```
LeanderK, line 9: why no interface?

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/notification/SQLEmailNotificationPolicy.java
```diff
@@ -0,0 +1,64 @@
...
+        }
+    }
+
+    @Override
+    protected void send(Notification notification, Result<Record> token) {
+        StringBuilder message = new StringBuilder();
+        message.append(notification.getDescription());
+        message.append("\n");
+        for (Record record : token) {
+            message.append(record.toString());
```
LeanderK: this might be impractical. If the results is large you get lines and lines of sql-results. Also a default header would be nice! Otherwise I don't know if the receiver of the email knows whats going on ;) I personally don't pay much attention to the title of emails, so i would be a bit confused why somebody is emailing me data

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/notification/SQLEmailNotificationPolicy.java
```diff
@@ -0,0 +1,64 @@
...
+ * @version 1.0
+ */
+public class SQLEmailNotificationPolicy extends NotificationPolicy<Result<Record>> {
+    private MailSender mailSender;
+    private String receiver;
+    private NotificationOperations operation;
+
+
+    /**
+     * @param mailSender an implementation of the MailSender interfac
```
LeanderK: typo

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/notification/NotificationController.java
simonkorz, line 78: no, I just added it for completeness

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/notification/NotificationController.java
simonkorz: Youre right. I might change it to just do nothing if no handle is returned. Or should I throw an exception?

LeanderK: i would throw an IllegalArgumentException

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/notification/NotificationController.java
simonkorz: yep, false is better. I think I misunderstood the documentation

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/notification/NotificationPolicy.java
simonkorz, line 9: I didnt want to make it too complicated. And this is kind of a mixture between strategy pattern and template method pattern. Abstract because it assures that invoke does always what it has to do.

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/notification/SQLEmailNotificationPolicy.java
simonkorz: Ahhh yes I wanted a default header too, but I forgot. 
Actually I wanted to parse the result as such, but unfortunately that doesnt work for an unspecified record type.
I might add a limit to the number of records that are sent, or concatetenate the string if it gets too long

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/notification/Notification.java
simonkorz, line 12: yep, but this class has a method of use in fact. see thresholdPassed()

LeanderK, line 12: i mean storing the record as a private variable and exposing it via getDBInstance or something comparable.

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/notification/NotificationPolicy.java
LeanderK, line 9: i mean that this class could also be an interface without any change, the invoke could be a default method and you would gain a bit a bit flexibility without much effort

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/notification/SQLEmailNotificationPolicy.java
LeanderK: i would just limit it. what do you mean with parse the result as such?

simonkorz: The usage of the method `result.format()`

LeanderK: just out of curiosity, what happens when you call `result.format()`?

kelunik: I would also limit it to 10 results or so.

kelunik:
Functionality > Beauty. ;-)

LeanderK:
this statement ist debatable :D

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/notification/SQLEmailNotificationPolicy.java
simonkorz: result.format() should return a string containing all records in the result with "table look".
But in this case it returns an empty table, because it can't parse it for an unspecified record type

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/notification/NotificationPolicy.java
simonkorz, line 9: If you mean changing that into a functional interface, one would have to split the interface. From the docs: "Conceptually, a functional interface has exactly one abstract method"

LeanderK, line 9: no, just a normal interface, since you are defining more of an interface with the methods invoke (default method), check and send

simonkorz, line 9: okay I see. I honestly didn't know about the "default" keyword in interfaces. I changed it now, but I see no real benefit.
Now other packages can send notifications because the send method is no more protected. I dont think this should be public

LeanderK, line 9: this is an OOP question...is you class an abstraction or an contract? If you have a contract its an interface (like i am list, i have a get(index), remove(index) etc. method) or an abstraction, like a Thread. its your decision, it never meant forcing you to implement an interface, this was more of a question

kelunik:
Why is this now closed?

LeanderK:
because its not ready for merging yet, and there are a few open points. I would like a clean start when he is ready for merging.

PR #54, opened 2016-02-01
database: added DatabaseMaintainer and fixed bug

LeanderK:
- added `DatabaseMaintainer`, used for cleaning up the db
- fixed a bug which was responsible for the experiments platforms not being deleted


LeanderK:
there is a task left on my list....

PR #55, opened 2016-02-01
Database

LeanderK:
- added `DatabaseMaintainer`which performs cleanup-tasks
- fixed a bug which caused the records from the ExperimentPlatforms to never being deleted
- added the automatic Experiment-Calibration


src/main/java/edu/kit/ipd/crowdcontrol/objectservice/database/DatabaseMaintainer.java
```diff
@@ -0,0 +1,69 @@
...
+        this.create = create;
+        this.cleanUpInterval = cleanUpInterval;
+    }
+
+    /**
+     * starts the DatabaseMaintainer
+     */
+    public synchronized void start() {
+        if (schedule != null) {
+            schedule = scheduler.schedule(this::doCleanUp, cleanUpInterval, TimeUnit.HOURS);
```
simonkorz: this should be: scheduler.scheduleAtFixedRate(this::doCleanUp, 0, cleanUpInterval, TimeUnit.HOURS);

LeanderK: thats right! thanks!

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/database/operations/AnswerRatingOperations.java
```diff
@@ -45,13 +56,31 @@ public AnswerRecord insertNewAnswer(AnswerRecord answerRecord) {
...
                         .insertInto(RATING)
                         .set(ratingRecord)
                         .returning()
                         .fetchOne()
         );
+        addToExperimentCalibration(ratingRecord.getWorkerId(), ratingRecord.getExperiment());
+        return result;
+    }
+
+    private void addToExperimentCalibration(int workerID, int experimentId) {
```
simonkorz: is this the "worker has answered task xy calibration"?

LeanderK: yes...and no ;) Every experiment has a calibration associated with it, where every worker working on an experiment can gets associated with (answered the the only question of the calibration). You can then exclude the workers which have previously worked on an experiment via the same mechanism as you would exclude a worker with the wrong calibration. (the automatic calibration gets excluded in the normal get calibrations-call). It's reducing the complexity in the database, os and the workerservice. Maybe i should write some javadoc!

simonkorz: Okay :) Javadoc is a good idea!

LeanderK:
push @phimonk

PR #56, opened 2016-02-01
Moneytransfer

TheRittler:
Newsletter package adjustments will follow in a different branch


TheRittler:
what to do about this fileNotFoundException? Add the gmailLogin.properties to git? (what @kelunik didnt want to)

TheRittler:
ok, nvm

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/database/operations/AbstractOperations.java
```diff
@@ -148,7 +148,7 @@ protected void assertHasField(MessageOrBuilder messageOrBuilder, int... fields)
      * @see #getNextRange(SelectWhereStep, Field, Table, Object, boolean, int, Comparator)
```
marcelhollerbach, line 161: I dont know how we proceed with shitespace changes here - but they are very very useless and annoying.

src/integration-test/java/edu/kit/ipd/crowdcontrol/objectservice/mail/MailHandlerTest.java
```diff
@@ -20,7 +20,7 @@
     protected String folder;
```
marcelhollerbach, line 20: why changing the mailhandler test in the moneytransfer pr ?

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/moneytransfer/AmazonMailFormatChangedException.java
```diff
@@ -0,0 +1,12 @@
...
+/**
+ * An exception, that gets thrown, if the parser cannot parse the mails, because the mail format changed.
+ * @author Felix Rittler
+ */
+public class AmazonMailFormatChangedException extends Exception{
+
+    public AmazonMailFormatChangedException() {
+        super("The Parser cannot extract the giftcode from the mails, because the mail format changed. You need to adjust the parser to the new mail format.");
+    }
+}
```
marcelhollerbach: Well if I get a amazon giftcard why not just caching the html text and send them again to the worker ? without parsing out a giftcode ? as far as I can see the cards are having a fancy link where you click on - log into your amazon acc. - and you got more money. Why parsing such a mail?

src/main/resources/db.sql
```diff
@@ -684,4 +684,4 @@ BEGIN
 END$$
 
 
-DELIMITER ;
+DELIMITER ;
```
marcelhollerbach: eeer whitespace change ? I think @LeanderK doesnt like that ...

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/database/operations/AbstractOperations.java
LeanderK, line 161: i also think they are a bit annoying

src/main/resources/db.sql
LeanderK: deleted the newline ;)

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/moneytransfer/AmazonMailFormatChangedException.java
TheRittler: Because, when we get a mail, we cant see the value of the GiftCode. And we need to log which worker gets which giftcode.

src/integration-test/java/edu/kit/ipd/crowdcontrol/objectservice/mail/MailHandlerTest.java
TheRittler, line 20: oops, this wasnt what i wanted to do ;)

LeanderK:
we need somebody to review the pull-request! @marcelhollerbach @kelunik @phimonk @lucaskrauss

marcelhollerbach:
I am doing it,tomorrow :P

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/moneytransfer/MailParser.java
```diff
@@ -0,0 +1,74 @@
...
+/**
+ * Describes a parser to extract amazon giftcodes from mails.
+ */
+public class MailParser {
+
+    protected static GiftCodeRecord parseAmazonGiftCode(Message msg) throws MessagingException, IOException, AmazonMailFormatChangedException{
+        //Extract Message
+        String message;
+        try {
+            Multipart parts = (Multipart) msg.getContent();
```
marcelhollerbach, line 37: does this really works ? isnt it a plain/text message?

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/moneytransfer/MailParser.java
```diff
@@ -0,0 +1,74 @@
...
+            throw new AmazonMailFormatChangedException();
+        }
+        //Parse Message
+        String messageStr = message.replaceAll(" ","");
+
+        String codePatternStr = "[0-9A-Z]+(-[0-9A-Z]+)+";
+        Pattern codePattern = Pattern.compile(codePatternStr);
+        Matcher codeMatcher = codePattern.matcher(messageStr);
+
+        String amountPatternStr = "[0-9]+[,|.][0-9][0-9][€]";
```
marcelhollerbach: what if someone adds 10€ ? Is this usecase somewhere documented ?

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/moneytransfer/MailParser.java
```diff
@@ -0,0 +1,74 @@
...
+        rec.setAmount(Integer.parseInt(amountStr));
+        rec.setCode(giftCode);
+
+        return rec;
+    }
+
+    private static boolean checkMatches(Matcher matcher) {
+        boolean match = true;
+
+        if (matcher.find()) {
```
marcelhollerbach, line 85: can you elaborate a bit more on how this function works ? or what it is doing ?

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/moneytransfer/MoneyTransferManager.java
```diff
@@ -0,0 +1,172 @@
...
+        this.workerOperations = workerOperations;
+        this.workerBalanceOperations = workerBalanceOperations;
+        this.minGiftCodesCount = 10;
+        this.payOffThreshold = 0;
+        this.notificationMailAddress = notificationMailAddress;
+        this.notificationText = new StringBuilder();
+    }
+
+    /**
+     * Logs a new money transfer and saves it.
```
marcelhollerbach, line 99: What does Log mean here ? is it really just a log?

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/moneytransfer/MoneyTransferManager.java
```diff
@@ -0,0 +1,172 @@
...
+     *
+     * @param workerID the id of the worker, who gets the money
+     * @param amount   the amount of money in ct
+     */
+    public void logMoneyTransfer(int workerID, int amount, int expID) {
+        workerBalanceOperations.addCredit(workerID, amount, expID);
+    }
+
+    /**
+     * Pays all workers depending on their logged money transfers.
```
marcelhollerbach: Okay looks like the entrys from above are getting payed here, maybe log is here a bit missleading, since most of the people will think of a Log like log-file where you just make a message and forget it.

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/moneytransfer/MoneyTransferManager.java
```diff
@@ -0,0 +1,172 @@
...
+     */
+    public void payOff() {
+        fetchNewGiftCodes();
+
+        Result<WorkerRecord> workers = workerOperations.getWorkerWithCreditBalanceGreaterOrEqual(payOffThreshold);
+        Iterator<WorkerRecord> workerIt = workers.iterator();
+        List<GiftCodeRecord> giftCodes = workerBalanceOperations.getUnusedGiftCodes();
+
+        while (workerIt.hasNext()) {
```
marcelhollerbach: I think a Result is iteratorable, this means you can go by 

```
for (WorkerRecord workerRecord : workers) {
...
}
```

So you dont have to get a iterator AND have a while.

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/moneytransfer/MoneyTransferManager.java
```diff
@@ -0,0 +1,172 @@
...
+    int minGiftCodesCount;
+    String notificationMailAddress;
+    StringBuilder notificationText;
+
+
+    public MoneyTransferManager(MailHandler mailHandler, WorkerBalanceOperations workerBalanceOperations, WorkerOperations workerOperations, String notificationMailAddress) throws MessagingException {
+        this.mailHandler = mailHandler;
+        this.workerOperations = workerOperations;
+        this.workerBalanceOperations = workerBalanceOperations;
+        this.minGiftCodesCount = 10;
```
marcelhollerbach: eeeeer I dont like that at all! maybe some dynamic cacluation ? lets say you have 10 times 1€, this is enough for a few weeks (under the assumation that only 10 workers with 1€ payload appear). What about calulation the number of outgoing codes from the last day + a few , and a minimum of 10 ? Or capture if a experiment is in creative stop, and if it is prepare to have enough money (optimistic caclulation since you dont know the bonuses)

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/moneytransfer/MoneyTransferManager.java
```diff
@@ -0,0 +1,172 @@
...
+                } catch (MessagingException f) {
+                    f.printStackTrace();
+                }
+                notificationText.append("It seems, that amazon changed the format of the giftcode E-Mails. You need to adjust the parser to import giftcodes in future.");
+            }
+        }
+    }
+
+    private List<GiftCodeRecord> chooseGiftCodes(WorkerRecord worker, List<GiftCodeRecord> giftCodes) {
+        List<GiftCodeRecord> payedCodes = new LinkedList<>();
```
marcelhollerbach: @LeanderK advise - never use a linked one ...

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/moneytransfer/MoneyTransferManager.java
```diff
@@ -0,0 +1,172 @@
...
+        int creditBalance = workerBalanceOperations.getBalance(worker.getIdWorker());
+        Iterator<GiftCodeRecord> giftCodesIt = giftCodes.iterator();
+
+        while (giftCodesIt.hasNext()) {
```
marcelhollerbach: A list is definitly interatable you dont need that, see the code snipped above.

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/moneytransfer/MoneyTransferManager.java
```diff
@@ -0,0 +1,172 @@
...
+     * @param amount   the amount of money in ct
+     */
+    public void logMoneyTransfer(int workerID, int amount, int expID) {
+        workerBalanceOperations.addCredit(workerID, amount, expID);
+    }
+
+    /**
+     * Pays all workers depending on their logged money transfers.
+     */
+    public void payOff() {
```
marcelhollerbach: Where is this called?

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/moneytransfer/MoneyTransferManager.java
```diff
@@ -0,0 +1,172 @@
...
+            while ((messageLine = reader.readLine()) != null) {
+                content.append(messageLine);
+                content.append(System.getProperty("line.separator"));
+            }
+        } catch (IOException e) {
+            e.printStackTrace();
+        }
+        return content;
+    }
+}
```
marcelhollerbach, line 362: A more general comment to this file, try to think if it is a good idea to print Stack Traces all over the place, or continue a Exception to a higher position in the Backtrace, so maybe the one who called payOff can handle what should happen next. And the output of the service is not that verbose when something fucked up

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/moneytransfer/MoneyTransferManager.java
marcelhollerbach: And this should maybe be another name...

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/moneytransfer/MailParser.java
TheRittler, line 37: no, it isnt. It is exactly a multipart message, that contains a multipart message, that contains a text/html part and a text/plain part (this is the part, which contains the giftcode, amount etc.)

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/moneytransfer/MailParser.java
TheRittler: you can add all amounts of money between 1 and MAXINTEGER

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/moneytransfer/MailParser.java
TheRittler, line 85: it checks that all expressions matching one of the regexes above in the mail have to be the same. So it cannot happen, that two strings, that seem to be a giftcode are in the mail. This would be an error.

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/moneytransfer/MoneyTransferManager.java
TheRittler: prof boehm wanted to set this minimum amount on his own, so i will outsource this to a config file, when/if we have one

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/moneytransfer/MailParser.java
marcelhollerbach, line 37: ok

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/moneytransfer/MailParser.java
marcelhollerbach: yeah sorry, missed the +

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/moneytransfer/MoneyTransferManager.java
TheRittler, line 99: it writes an outstanding balance to the db. You could use it as a synonym of record. The payment happens in the payOff() method

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/moneytransfer/MailParser.java
marcelhollerbach, line 85: this was meant as "can you write that into the code". The more verbose you are with comments the easier it is for people who are reviewing

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/moneytransfer/MoneyTransferManager.java
TheRittler: oh yeah, I forgot to change that

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/moneytransfer/MoneyTransferManager.java
marcelhollerbach: 1. I think there is one
2. Sure he wanted to have that ? (I think we should discuss that in private)

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/moneytransfer/MoneyTransferManager.java
TheRittler, line 362: the only thing, that can happen here is when the file is missing or unreadable, and this is a mistake of the serveradmin, because normally the file is there^^ So what to do?

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/moneytransfer/MoneyTransferManager.java
TheRittler: why another name? 
Actually this is called nowhere, at the state of draft we discussed about a future object triggering off payOff()

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/moneytransfer/MailParser.java
TheRittler, line 85: okay ;)

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/moneytransfer/MoneyTransferManager.java
marcelhollerbach, line 362: I was talking about the complete file. What to do would be for example in payWorker, return that the payment failed, and write back that the worker did not get any money so the action ca be triggered again. Something like that increases the code quality - and makes reusing the code easier.

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/moneytransfer/MoneyTransferManager.java
marcelhollerbach: well its a method which is doing payment of a bunch of workers - its way more than just "payOff"

Btw. at this function again return type void. What if this payOff would be called form a rest call - which has failed for the case of an exception - how would you recognize that there was a error?

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/moneytransfer/MoneyTransferManager.java
marcelhollerbach, line 99: ok - as I said below, using log in this context is a bit dangerous.

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/moneytransfer/MoneyTransferManager.java
LeanderK: yeah, they are only useful if you have extremely large lists and you are only adding/removing on head and tail. But then you should most likely use a queue.

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/moneytransfer/MoneyTransferManager.java
TheRittler: how would you name it? payOffAllWorkers? I would say, the rest is is implementation detail
At the moment it is not called by a rest call.

marcelhollerbach: submitGiftCodes() would be better I guess.

Let me be a bit more precisely, this here is payment stuff - every error should be given back to the caller, so he knows what is happening why things are failing. Lets assume you want to pay a worker and one emails fails to send - for the caller of this function everything went fine - which is not true at all. More clear: stop just printing stacktraces all over the place, redirect exceptions, with detailed messages what exactly is happening. Maybe create a new one so you dont throw 5 different exceptions.

LeanderK:
i mean it's up to you, guys, but i would just close the pull request, resolve the issues and then reopen it. I think it's easier to understand where a discussion is going on and where deficits are. Also as an reviewer it's not that easy to notice when the issues are resolved. especially if there is a long discussion and many commits.

LeanderK:
especially if there is a long discussion and many commits.

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/moneytransfer/MoneyTransferManager.java
```diff
@@ -0,0 +1,255 @@
...
+    private int payOffThreshold;
+    private int minGiftCodesCount;
+
+    private String notificationMailAddress;
+    private StringBuilder notificationText;
+
+    private ScheduledExecutorService scheduler = Executors.newScheduledThreadPool(1);
+    private ScheduledFuture<?> schedule = null;
+
+    private static final Logger LOGGER = LogManager.getRootLogger();
```
kelunik: Should be: `private static final Logger LOGGER = LogManager.getLogger(MoneyTransferManager.class);`

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/moneytransfer/MailParser.java
```diff
@@ -0,0 +1,81 @@
...
+
+        GiftCodeRecord rec = new GiftCodeRecord();
+        rec.setAmount(Integer.parseInt(amountStr));
+        rec.setCode(giftCode);
+
+        return rec;
+    }
+
+
+    private static boolean checkMatches(Matcher matcher) {
```
marcelhollerbach, line 82: I think your are going to hate me now, but why not writing it as javadoc above the method like everywhere else? 
Just as comment.

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/moneytransfer/MoneyTransferManager.java
```diff
@@ -0,0 +1,262 @@
...
+        } catch (UnsupportedEncodingException e) {
+            LOGGER.error(e);
+        }
+    }
+
+    private void sendErrorMessage(String errorMessage) {
+        StringBuilder message = new StringBuilder();
+        message.append("Dear administrator, ").append(System.getProperty("line.separator"));
+        message.append("it seems, that a critical error during payment occured:").append(System.getProperty("line.separator"));
+        message.append(errorMessage);
```
marcelhollerbach: Can you also move this message into a text file? So for example a translater could simply translate the message for the admin.

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/moneytransfer/MoneyTransferManager.java
```diff
@@ -0,0 +1,262 @@
...
+            }
+        }
+    }
+
+    private void sendNotification() throws MoneyTransferException {
+        StringBuilder message = new StringBuilder();
+        message.append("Dear administrator, ").append(System.getProperty("line.separator"));
+        message.append("we want to give you the following information:").append(System.getProperty("line.separator"));
+        message.append(notificationText);
+
```
marcelhollerbach, line 294: Same as below

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/moneytransfer/MoneyTransferManager.java
```diff
@@ -0,0 +1,262 @@
...
+            paymentMessage = new StringBuilder(Template.apply(paymentMessage.toString(), map));
+
+            //sends payment message
+            try {
+                mailHandler.sendMail(worker.getEmail(), "Your payment for your Crowdworking", paymentMessage.toString());
+            } catch (MessagingException e) {
+                throw new MoneyTransferException("The MailHandler couldnt send mails to crowdworkers." +
+                        "It seems, that there is either a problem with the server or with the properties file.");
+            } catch (UnsupportedEncodingException e) {
+                LOGGER.error(e);
```
marcelhollerbach: For the cases that the messaging failed, shouldnt the gift code be recovered ? So he can be used again?

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/moneytransfer/MoneyTransferManager.java
```diff
@@ -0,0 +1,262 @@
...
+            }
+        }
+    }
+
+    private void sendNotification() throws MoneyTransferException {
```
marcelhollerbach: also why not using a parameter like in send Error Message ? 

And maybe even further - make one sendLog(String msg, String urgency) where you send a email with this urgency and message. would cleanup the code a bit.

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/moneytransfer/MoneyTransferManager.java
```diff
@@ -0,0 +1,262 @@
...
+        }
+
+        if (creditBalance >= payOffThreshold && creditBalance >= 15 && !payedCodes.isEmpty()) {
+            notificationText = notificationText.append("A worker has pending Payments in the amount of ").append(creditBalance).append("ct. Please add giftcodes, so the payment of the worker can be continued.").append(System.getProperty("line.separator"));
+        }
+        return payedCodes;
+    }
+
+    private void payWorker(WorkerRecord worker, List<GiftCodeRecord> giftCodes) throws MoneyTransferException {
+        if (!giftCodes.isEmpty()) {
```
marcelhollerbach: if gift codes are empty we are just doing nothing ? shouldnt we loadly complain "the world is on fire" ?

marcelhollerbach: never mind ... wrong function...

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/moneytransfer/MailParser.java
TheRittler, line 82: bc its a private methode, but if you want that...

marcelhollerbach, line 82: Oh if you ask like that, I would like to have detailed javadoc on each and every function, 

Just a question for yourself, step back and try to read your own code with ignoring the knowledge of that you wrote the code, what is the function checkmatches doing? So you see not only commenting on public functions is usefull, also private ones.

The biggest reason for writing it as javadoc is are ide's most of them are showing javadoc on a function when you mouseover on sm. Functio, so its also useful here.

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/moneytransfer/MoneyTransferManager.java
TheRittler: you mean merge sendNotificationMessage() and sendErrorMessage()?

marcelhollerbach: Yeah

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/moneytransfer/MoneyTransferManager.java
TheRittler: is fixed. Now giftcodes get marked as used directly after sending

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/moneytransfer/MoneyTransferManager.java
```diff
@@ -0,0 +1,293 @@
...
+            StringBuilder paymentMessage = loadMessage("src/main/resources/PaymentMessage.txt");
+            StringBuilder giftCodeMessage = new StringBuilder();
+
+            //saves payment to the database
+            for (GiftCodeRecord rec : giftCodes) {
+                giftCodeMessage.append(rec.getCode()).append(System.getProperty("line.separator"));
+            }
+
```
simonkorz, line 259: where is anything saved here? i don't understand the comment

TheRittler, line 259: this comment is outdated, i'll change it

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/moneytransfer/MoneyTransferManager.java
```diff
@@ -0,0 +1,292 @@
...
+        LOGGER.trace("Started submission of giftcodes to workers.");
+        Result<WorkerRecord> workers = workerOperations.getWorkerWithCreditBalanceGreaterOrEqual(payOffThreshold);
+        List<GiftCodeRecord> giftCodes = workerBalanceOperations.getUnusedGiftCodes();
+
+        for (WorkerRecord worker : workers) {
+            if (!worker.getEmail().equals("")) {
+                List<GiftCodeRecord> payedCodesForWorker = chooseGiftCodes(worker, giftCodes);
+
+                giftCodes = workerBalanceOperations.getUnusedGiftCodes();
```
marcelhollerbach: shouldnt this happen _after_ payWorker ? i cannot see where in chooseGiftCodes the codes are getting marked used ...

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/moneytransfer/MoneyTransferManager.java
```diff
@@ -0,0 +1,292 @@
...
+        } catch (MessagingException e) {
+            throw new MoneyTransferException("The MailHandler couldn't fetch new giftcodes from the mailserver. " +
+                    "It seems, that there is either a problem with the server or with the properties file.");
+        }
+
+        //extract giftcodes and save them to the database
+        for (Message message : messages) {
+            try {
+                GiftCodeRecord rec = MailParser.parseAmazonGiftCode(message);
+                if (rec != null) {
```
marcelhollerbach: is there really the case that null can be returned ? shoudnt the function always return rec or throw up ?

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/moneytransfer/MoneyTransferManager.java
```diff
@@ -0,0 +1,292 @@
...
+                    workerBalanceOperations.addGiftCode(rec.getCode(), rec.getAmount());
+                }
+            } catch (MoneyTransferException e) {
+                try {
+                    mailHandler.markAsUnseen(message);
```
marcelhollerbach, line 170: so this means that a mail will be parsed again at the next payment iteration and fail again?

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/moneytransfer/MoneyTransferManager.java
TheRittler: IT Returns null, if the Massage is not from Amazon

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/moneytransfer/MoneyTransferManager.java
TheRittler, line 170: This should happen this way, because it just throws an exception, if the mail is an amazon mail, and cannot get parsed (e. G. The format changed) so the mail should get read the next time too, or the time when the parser was uodated

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/moneytransfer/MailParser.java
```diff
@@ -0,0 +1,99 @@
...
+     *
+     * @param msg the message, to extract the code
+     * @return returns the giftcode
+     * @throws MoneyTransferException gets thrown, if an error occurred during parsing
+     */
+    protected static GiftCodeRecord parseAmazonGiftCode(Message msg) throws MoneyTransferException {
+        //Extract Message
+        String message;
+        try {
+            if (!msg.getFrom()[0].toString().toLowerCase().contains("amazon.de")) {
```
marcelhollerbach: amazon.de@t-online.de is also valid here, maybe you need a better checking here.

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/database/operations/WorkerBalanceOperations.java
```diff
@@ -70,15 +70,13 @@ public boolean addCredit(int workerID, int amount, int experiment) {
      * adds a debit to the balance-sheet of the workers wages.
      * @param workerID the primary key of the worker the debit belongs to
      * @param amount the amount that got payed
```
kelunik, line 72: Should have information that it's negative, because mathematical amount is always positive.

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/moneytransfer/MailParser.java
```diff
@@ -0,0 +1,101 @@
...
+     * @param msg the message, to extract the code
+     * @return returns the giftcode
+     * @throws MoneyTransferException gets thrown, if an error occurred during parsing
+     */
+    protected static Optional<GiftCodeRecord> parseAmazonGiftCode(Message msg) throws MoneyTransferException {
+        //Extract Message
+        String message;
+        Optional<GiftCodeRecord> optional;
+        try {
+            if (!msg.getFrom()[0].toString().toLowerCase().endsWith("amazon.de>")) {
```
kelunik: Probably unsafe, `getFrom` might also be just a simple mail address without `Description <address>`.

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/moneytransfer/MailParser.java
```diff
@@ -0,0 +1,101 @@
...
+                return Optional.ofNullable(null);
```
kelunik: Shouldn't this be an exception?

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/moneytransfer/MailParser.java
```diff
@@ -0,0 +1,101 @@
...
+            }
+        } catch (MessagingException e) {
+            return Optional.ofNullable(null);
```
kelunik: Shouldn't this be an exception?

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/moneytransfer/MailParser.java
```diff
@@ -0,0 +1,101 @@
...
+        }
+        try {
+            Multipart parts = (Multipart) msg.getContent();
+            BodyPart body = parts.getBodyPart(0);
+            Multipart innerMsg = (Multipart) body.getContent();
+            BodyPart textBody = innerMsg.getBodyPart(0);
+            message = textBody.getContent().toString();
+        } catch (ClassCastException | MessagingException | IOException e) {
+            throw new MoneyTransferException("The Parser cannot extract the giftcode from the mails, because the mail format changed. You need to adjust the parser to the new mail format.");
```
kelunik, line 43: Should probably log the `messageId`. Every mail _should_ have a unique one if it's not spam.

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/moneytransfer/MailParser.java
```diff
@@ -0,0 +1,101 @@
...
+        }
+        //Parse Message
+        String messageStr = message.replaceAll(" ", "");
+
+        StringBuilder password = MoneyTransferManager.loadMessage("src/main/resources/parsingPassword.txt");
```
kelunik: Should probably injected in the constructor and be in the main config.

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/moneytransfer/MailParser.java
kelunik: `[,|.]`, why's there a `|` in there? It's a character class, it will match `|`. I don't think that's intended.

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/moneytransfer/MailParser.java
```diff
@@ -0,0 +1,101 @@
...
+        return Optional.of(rec);
+    }
+
+    /**
+     * Checks, whether all groups of a matcher are identical.
+     *
+     * @param matcher the matcher to check
+     * @return returns, whether the groups are identical
+     */
+    private static boolean checkMatches(Matcher matcher) {
```
kelunik, line 82: What's the purpose of this method?

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/moneytransfer/MoneyTransferManager.java
```diff
@@ -0,0 +1,291 @@
...
+
+    private static final Logger LOGGER = LogManager.getLogger(MoneyTransferManager.class);
+    private MailHandler mailHandler;
+    private WorkerBalanceOperations workerBalanceOperations;
+    private WorkerOperations workerOperations;
+    private int payOffThreshold;
+    private String notificationMailAddress;
+    private StringBuilder notificationText;
+    private ScheduledExecutorService scheduler = Executors.newScheduledThreadPool(1);
+    private ScheduledFuture<?> schedule = null;
```
kelunik, line 31: Most of those fields can probably be `final`.

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/moneytransfer/MoneyTransferManager.java
```diff
@@ -0,0 +1,291 @@
...
+                    sendNotification(e.toString(), true);
+                } catch (MoneyTransferException f) {
+                    LOGGER.error(f);
+                }
+                LOGGER.error(e);
+            }
+        };
+
+
+        schedule = scheduler.scheduleAtFixedRate(runnable, 7, 7, TimeUnit.DAYS);
```
kelunik: `7` should probably a constant in the class, so it's easy to find and change. Something like `SCHEDULE_INTERVAL_DAYS`

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/moneytransfer/MoneyTransferManager.java
```diff
@@ -0,0 +1,291 @@
...
+     * @param mailHandler             the mailhandler, used to fetch new giftcodes and send notification and payment messages
+     * @param workerBalanceOperations the workerBalanceOperations, used to change the credit balance of a worker
+     * @param workerOperations        the workerOperations, used to do operations on workers
+     * @param notificationMailAddress the mail address to send notifications
+     */
+    public MoneyTransferManager(MailHandler mailHandler, WorkerBalanceOperations workerBalanceOperations, WorkerOperations workerOperations, String notificationMailAddress) {
+        this.mailHandler = mailHandler;
+        this.workerOperations = workerOperations;
+        this.workerBalanceOperations = workerBalanceOperations;
+        this.payOffThreshold = 0;
```
kelunik: Constant? Should be uppercase and final.

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/moneytransfer/MoneyTransferManager.java
```diff
@@ -0,0 +1,291 @@
...
+ */
+public class MoneyTransferManager {
+
+    private static final Logger LOGGER = LogManager.getLogger(MoneyTransferManager.class);
+    private MailHandler mailHandler;
+    private WorkerBalanceOperations workerBalanceOperations;
+    private WorkerOperations workerOperations;
+    private int payOffThreshold;
+    private String notificationMailAddress;
+    private StringBuilder notificationText;
```
kelunik: This must not be a class global object. Create a new local one for each worker.

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/moneytransfer/MoneyTransferManager.java
```diff
@@ -0,0 +1,291 @@
...
+        //fetch new mails
+        try {
+            messages = mailHandler.fetchUnseen("inbox");
+        } catch (MessagingException e) {
+            throw new MoneyTransferException("The MailHandler couldn't fetch new giftcodes from the mailserver. " +
+                    "It seems, that there is either a problem with the server or with the properties file.");
+        }
+
+        //extract giftcodes and save them to the database
+        for (Message message : messages) {
```
kelunik, line 161: Please log all `messageIds` as `INFO`. Might also save them in the DB and keep our own `unseen` state instead of trusting the inbox.

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/moneytransfer/MoneyTransferManager.java
```diff
@@ -0,0 +1,291 @@
...
+    }
+
+    /**
+     * Sends a notification about problems or errors during payment to the administrator.
+     *
+     * @param message the message to send
+     * @param criticalError the urgency of the problem
+     * @throws MoneyTransferException gets thrown, if an error occurred
+     */
+    private void sendNotification(String message, boolean criticalError) throws MoneyTransferException {
```
kelunik: I would split this into two methods. Those can use a single one that gets the subject passed. But a `boolean` in a method signature is always fishy.

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/moneytransfer/MoneyTransferManager.java
```diff
@@ -0,0 +1,291 @@
...
+        StringBuilder mail;
+        String subject = "";
+        try {
+            if (criticalError) {
+                subject = "Payment Error occurred";
+                mail = loadMessage("src/main/resources/errorMoneyTransferMessage.txt");
```
kelunik: Maybe put them into a sub directory to keep things clean.

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/moneytransfer/MoneyTransferManager.java
```diff
@@ -0,0 +1,291 @@
...
+
+        mail = mail.append(message);
+        try {
+            if (notificationText.length() != 0) {
+                LOGGER.trace("Started sending a notification about problems or errors with submission of giftcodes.");
+                mailHandler.sendMail(notificationMailAddress, subject, mail.toString());
+                LOGGER.trace("Completed sending a notification about problems or errors with submission of giftcodes.");
+            }
+        } catch (MessagingException e) {
+            throw new MoneyTransferException("The MailHandler couldnt send mails to the administrator." +
```
kelunik: This error message is often used. Maybe refactor into a constant?

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/moneytransfer/MoneyTransferManager.java
```diff
@@ -0,0 +1,291 @@
...
+
+
+    /**
+     * Loads a string from a file.
+     *
+     * @param path the path to the file
+     * @return the StringBuilder to return
+     * @throws MoneyTransferException gets thrown, if an error occurred
+     */
+    protected static StringBuilder loadMessage(String path) throws MoneyTransferException {
```
kelunik: Should probably return a `String` instead of the `StringBuilder` object.

src/main/resources/parsingPassword.txt
```diff
@@ -0,0 +1 @@
+test
```
kelunik: Must be in `.gitignore` and should be in global config instead, as already mentioned.

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/database/operations/WorkerBalanceOperations.java
LeanderK, line 72: i thinks its an error, it should be positive and the flipped in (my :D ) addDebit method. I would change it when @redalertexpert is ready.

TheRittler, line 72: not my package, but i think if i give as param a positive amount, this amount gets subtracted (thats the way i used this method)

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/moneytransfer/MailParser.java
```diff
@@ -0,0 +1,101 @@
...
+/**
+ * Describes a parser to extract amazon giftcodes from mails.
+ */
+public class MailParser {
+
+    /**
+     * Extracts a giftcode out of a message.
+     *
+     * @param msg the message, to extract the code
+     * @return returns the giftcode
```
LeanderK: when does it return empty?

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/database/operations/WorkerBalanceOperations.java
LeanderK, line 72: ok good

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/moneytransfer/MailParser.java
TheRittler: then i should use a pattern

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/moneytransfer/MailParser.java
TheRittler: no, null are mails, that weren't sent by amazon, but other mails

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/moneytransfer/MailParser.java
TheRittler: look at my comment above

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/moneytransfer/MailParser.java
TheRittler, line 43: Why?

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/moneytransfer/MailParser.java
kelunik: Just check additionally for the version without `>`.

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/moneytransfer/MailParser.java
TheRittler: ok

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/moneytransfer/MailParser.java
TheRittler, line 82: additional check, that i dont have more than 1 string in the mail, that match my giftcode/amount pattern, so i dont save wrong giftcodes.

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/moneytransfer/MailParser.java
kelunik, line 43: Because we're dealing with money and everything should as comprehensible as possible.

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/moneytransfer/MoneyTransferManager.java
TheRittler, line 31: you mean the scheduler? And which others?

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/moneytransfer/MoneyTransferManager.java
TheRittler: ok

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/moneytransfer/MailParser.java
kelunik, line 82: Does every mail contain just one giftcode?

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/moneytransfer/MoneyTransferManager.java
TheRittler: i think i will remove this at all

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/moneytransfer/MoneyTransferManager.java
TheRittler: then i will need a new one in each submitGiftCodes() call?

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/database/operations/WorkerBalanceOperations.java
LeanderK, line 72: resolved

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/moneytransfer/MailParser.java
```diff
@@ -0,0 +1,101 @@
...
+     * @throws MoneyTransferException gets thrown, if an error occurred during parsing
+     */
+    protected static Optional<GiftCodeRecord> parseAmazonGiftCode(Message msg) throws MoneyTransferException {
```
LeanderK: you don't use the Optional! you always return something

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/moneytransfer/MoneyTransferManager.java
kelunik, line 31: `mailHandler`, `workerBalanceOperations`, `workerOperations`, `payOffThreshold`, `notificationMailAddress`.

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/moneytransfer/MoneyTransferManager.java
kelunik: Yup.

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/moneytransfer/MoneyTransferManager.java
```diff
@@ -0,0 +1,291 @@
...
+        this.notificationText = new StringBuilder();
+    }
+
+    /**
+     * Starts the MoneyTransferManager, so giftcodes become submitted to workers every 7 days.
+     */
+    public synchronized void start() {
+        if (schedule != null) {
+            throw new IllegalStateException("run() was called twice!");
+        }
```
LeanderK: this is not thread-safe!

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/moneytransfer/MailParser.java
LeanderK: you should use optional.emtpy()

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/moneytransfer/MailParser.java
LeanderK: you should use optional.emtpy()

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/moneytransfer/MoneyTransferManager.java
```diff
@@ -0,0 +1,291 @@
...
+    }
+
+    /**
+     * Pays all workers depending on their logged money transfers.
+     * @throws MoneyTransferException gets thrown, if an error occurred
+     */
+    public void submitGiftCodes() throws MoneyTransferException {
+        LOGGER.trace("Started fetching new giftcodes.");
+        fetchNewGiftCodes();
+        LOGGER.trace("Completed fetching new giftcodes.");
```
LeanderK: i would mention how many

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/moneytransfer/MoneyTransferManager.java
```diff
@@ -0,0 +1,291 @@
...
+        LOGGER.trace("Started submission of giftcodes to workers.");
+        Result<WorkerRecord> workers = workerOperations.getWorkerWithCreditBalanceGreaterOrEqual(payOffThreshold);
+        List<GiftCodeRecord> giftCodes = workerBalanceOperations.getUnusedGiftCodes();
+
+        for (WorkerRecord worker : workers) {
+            if (!worker.getEmail().equals("")) {
+                List<GiftCodeRecord> payedCodesForWorker = chooseGiftCodes(worker, giftCodes);
+
+                payWorker(worker, payedCodesForWorker);
+                giftCodes = workerBalanceOperations.getUnusedGiftCodes();
```
LeanderK, line 131: more logging, this is crucial.

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/moneytransfer/MoneyTransferManager.java
```diff
@@ -0,0 +1,291 @@
...
+
+        for (GiftCodeRecord nextCode : giftCodes) {
+            if (creditBalance == 0) {
+                break;
+            }
+            if (nextCode.getAmount() <= creditBalance) {
+                payedCodes.add(nextCode);
+                creditBalance -= nextCode.getAmount();
+            }
+        }
```
LeanderK, line 231: this is not a very smart algorithm. You should optimise to use the least possible gift-codes and achieve credit-balance = 0. Or you can optimise to pay the workers the max. amount possible. I would have been smart to ask Böhm what he want. But this is not a good solution.

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/moneytransfer/MoneyTransferManager.java
```diff
@@ -0,0 +1,291 @@
...
+
+    /**
+     * Builds the payment message and sends giftcodes to the worker.
+     *
+     * @param worker    the worker to pay
+     * @param giftCodes the giftcodes to send
+     * @throws MoneyTransferException gets thrown, if an error occured
+     */
+    private void payWorker(WorkerRecord worker, List<GiftCodeRecord> giftCodes) throws MoneyTransferException {
+        if (!giftCodes.isEmpty()) {
```
LeanderK: this is extremely ugly. use: `if (!giftCodes.isEmpty()) { return; }`

kelunik: `if (giftCodes.isEmpty()) { return; }` :stuck_out_tongue_winking_eye:

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/moneytransfer/MoneyTransferManager.java
TheRittler: @marcelhollerbach suggested this to me, i had 2 methods before...

marcelhollerbach: Ehm well, lets see the whole situation, you had 2 functions which did exactly the same beside a different level of log, one attached "Error" one attached "Information". I said they could be merged. Yes, I never said you should add this kind of boolean, You could have added something like sendNotification(String message, String level) or attach the log level before to the message. I never gave you this signiture.

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/moneytransfer/MailParser.java
redalertexpert: yes, but i dont get your point. what does the message id give us helpful? And where should i log it? We are already tracking incoming giftcodes and all payments to workers

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/moneytransfer/MailParser.java
TheRittler, line 82: yes. If you buy 10 giftcodes, you get 10 mails with codes

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/moneytransfer/MoneyTransferManager.java
TheRittler, line 31: do we have to restart the object-service, if someone changes the config file?

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/moneytransfer/MoneyTransferManager.java
TheRittler, line 161: what do you mean with "keep our own unseen state"?

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/moneytransfer/MoneyTransferManager.java
TheRittler: a String level is actually not needed. a boolean is completely sufficient. i can't recognize value in a string instead of a boolean.

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/moneytransfer/MoneyTransferManager.java
TheRittler: one moment: which object do you mean? the notification text? and why?

src/main/resources/parsingPassword.txt
TheRittler: when i put it in the global config, i can't put it in gitignore, bc i shouldn't put the global config in gitignore

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/moneytransfer/MoneyTransferManager.java
TheRittler, line 231: This is the only solution, all other problems are np-complete, or i don't get what you want.

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/moneytransfer/MoneyTransferManager.java
kelunik: A boolean is absolutely not what we want here. We want three methods:
- `sendMail(String subject, String body)`
- `sendCriticalNotification(String message)` (uses `sendMail`)
- `sendNotification(String message)` (uses `sendMail`)

Something like that.

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/moneytransfer/MailParser.java
TheRittler: ehm, for sure, thats not intended :D

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/moneytransfer/MoneyTransferManager.java
LeanderK, line 231: well being np-complete does not mean that is is not solvable o.O. Its just not polynomial...I also don't know if there are approximations. You don't have to guarantee that you solve the problem, but you also don't do this right now. Sry but this is not a very smart algorithm.

redalertexpert: the ONLY solution to guarantee that the payment is maximal is try out all possible combinations of giftcode values. I can do this, but this will take exponential time. And this becomes to a problem, if there are many giftcodes with different values in the database.

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/moneytransfer/MoneyTransferManager.java
```diff
@@ -0,0 +1,336 @@
...
+    private StringBuilder notificationText;
+    private String MAIL_FAILURE_MESSAGE = "The MailHandler was not able to send mails." +
+            "It seems, that there is either a problem with the server or with the properties file.";
+    private ScheduledFuture<?> schedule = null;
+
+    //config-file data
+    private final int PAY_OFF_THRESHOLD;
+    private final String PARSING_PASSWORD;
+    private final String NOTIFICATION_MAIL_ADDRESS;
+    private final int SCHEDULE_INTERVAL_DAYS;
```
kelunik: They should normal `camelCase`, use `UPPER_CASE` only when they're static.

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/moneytransfer/MoneyTransferManager.java
```diff
@@ -0,0 +1,336 @@
...
+public class MoneyTransferManager {
+
+    private static final Logger LOGGER = LogManager.getLogger(MoneyTransferManager.class);
+    private final MailHandler MAIL_HANDLER;
+    private final WorkerBalanceOperations WORKER_BALANCE_OPERATIONS;
+    private final WorkerOperations WORKER_OPERATIONS;
+    private final ScheduledExecutorService SCHEDULER = Executors.newScheduledThreadPool(1);
+
+    private StringBuilder notificationText;
+    private String MAIL_FAILURE_MESSAGE = "The MailHandler was not able to send mails." +
```
kelunik: `static`?

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/moneytransfer/MoneyTransferManager.java
```diff
@@ -0,0 +1,336 @@
...
+                break;
+            }
+            if (nextCode.getAmount() <= creditBalance) {
+                payedCodes.add(nextCode);
+                creditBalance -= nextCode.getAmount();
+            }
+        }
+
+        if (creditBalance >= PAY_OFF_THRESHOLD && creditBalance >= 15) {
+            notificationText = notificationText.append("A worker has pending Payments in the amount of ").append(creditBalance).append("ct. Please add giftcodes, so the payment of the worker can be continued.").append(System.getProperty("line.separator"));
```
kelunik: Why do you use `=` here?

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/moneytransfer/MoneyTransferManager.java
kelunik: Because it leads to issues once you invoke it twice. Actually, it already is an issue, because you never reset the builder anywhere.

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/moneytransfer/MoneyTransferManager.java
TheRittler: this was a String before, i didnt change it

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/moneytransfer/MoneyTransferManager.java
TheRittler, line 231: is fixed. i implemented an efficient verision with dynamic programming. Optimal solutions are now guaranteed.

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/moneytransfer/MoneyTransferManager.java
kelunik, line 131: _push_

TheRittler, line 131: is fixed

PR #65, opened 2016-02-03
Fix: Experiment API

kelunik:
- Return `templateId` correctly
- Return full experiments in experiment lists

### DIFF
--- a/src/main/java/edu/kit/ipd/crowdcontrol/objectservice/rest/resources/ExperimentResource.java
+++ b/src/main/java/edu/kit/ipd/crowdcontrol/objectservice/rest/resources/ExperimentResource.java
@@ -65,6 +65,7 @@ public class ExperimentResource {
         int from = getQueryInt(request, "from", 0);
         boolean asc = getQueryBool(request, "asc", true);
 
+        // TODO: (low priority) Optimize fetchExperiment for multiple experiments
         return experimentOperations.getExperimentsFrom(from, asc, 20)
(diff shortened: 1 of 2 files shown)


src/main/java/edu/kit/ipd/crowdcontrol/objectservice/rest/resources/ExperimentResource.java
```diff
@@ -66,10 +66,7 @@ public ExperimentResource(ExperimentOperations experimentOperations, Calibration
         boolean asc = getQueryBool(request, "asc", true);
 
         return experimentOperations.getExperimentsFrom(from, asc, 20)
-                .map(experimentRecord -> ExperimentTransformer.toProto(
-                        experimentRecord,
-                        experimentOperations.getExperimentState(experimentRecord.getIdExperiment()))
-                )
+                .map(experimentRecord -> fetchExperiment(experimentRecord.getIdExperiment()))
```
LeanderK, line 70: might leave a comment that you could very easily optimise this is you want to, it's just a bit of work

kelunik, line 70: Just a `// TODO: Optimize for performance`?

LeanderK, line 70: i would add a maybe, but thats a pretty low hanging fruit, so i would the TODO

PR #66, opened 2016-02-03
Proto encapsulate

kelunik:


LeanderK:
it would be cool if in the future you would explain a bit what happened in the protos, so the reviewer does not have to guess based on commit messages and code

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/database/transformers/ExperimentTransformer.java
```diff
@@ -16,19 +17,28 @@
...
+ * Handles transformations of experiments from and to the protocol buffer messages.
+ *
+ * @author Leander K.
  * @author Marcel Hollderbach
+ * @author Niklas Keller
  */
 public class ExperimentTransformer extends AbstractTransformer {
+    private static Integer toInteger(int value) {
+        return Integer.newBuilder().setValue(value).build();
+    }
```
LeanderK: i would move this in AbstractTransformer

kelunik:
I even linked the appropriate commit in the spec. What exactly should I explain here?

LeanderK:
no, in the description of the pull-request...its not that big of an issue ;)

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/rest/RequestUtil.java
```diff
@@ -5,6 +5,10 @@
 import spark.Request;
 
 public class RequestUtil {
+    private RequestUtil() {
+        // utility class
+    }
```
LeanderK: i would write some javadoc ;) just to justify the private constructor

kelunik: Usually documentation for private members isn't generated anyway and it's justified by the comment inside?

LeanderK: utility class are still not always static. And they are not generated when you export to html, but the IDE still shows them. But i was talking more about the class-javadoc. something like:
`\** bla bla contains static methods to help with Sparkjava-requests....`

kelunik: Oh, Javadoc on the class is actually miss, right. Added it.

PR #78, opened 2016-02-04
CrowdWorking: Type and dummy-platform

LeanderK:
- implemented `DummyPlatform` used for testing/exploring the api/verifying the worker-service is responsive
- implemented `FallbackWorker`
- added `getType` to platform
- added missing patch worker rest-command

### DIFF
--- a/src/main/java/edu/kit/ipd/crowdcontrol/objectservice/crowdworking/FallbackWorker.java
+++ b/src/main/java/edu/kit/ipd/crowdcontrol/objectservice/crowdworking/FallbackWorker.java
@@ -1,0 +1,35 @@
+package edu.kit.ipd.crowdcontrol.objectservice.crowdworking;
+
+import org.apache.commons.validator.routines.EmailValidator;
+
+import java.util.Map;
+
+/**
+ * The fallback-worker uses the email-address to identify workers.
+ * @author LeanderK
+ * @version 1.0
+ */
+public class FallbackWorker implements WorkerIdentification {
+    /**
+     * Parse a worker id out of the params
+     *
+     * @param param The parameters which were sent by a platform
+     * @return The id of the worker if one can be found
+     */
+    @Override
+    public String identifyWorker(Map<String, String[]> param) throws UnidentifiedWorkerException {
+        //if the worker provides an email-address return it
+        if (param.containsKey("email")) {
+            String email = param.get("email")[0];
+            if (!EmailValidator.getInstance(false).isValid(email)) {
+                throw new UnidentifiedWorkerException(String.format("invalid email: %s!", email));
+            }
+            return email;
+            //if the worker does not provide an email-address return not an email
+            // (this can not be found in our database, since our identification is an email.
+            // The worker then gets asked for an email).
+        } else {
+            return "_!";
+        }
+    }
+}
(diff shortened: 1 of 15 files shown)


src/main/java/edu/kit/ipd/crowdcontrol/objectservice/crowdworking/Platform.java
```diff
@@ -11,33 +11,42 @@
...
     String getName();
 
     /**
+     * Returns the type of the platform.
+     * <p>
+     * The String must match {@code [a-z0-9_]+} and must be a constant since it is the only way other services can
+     * communicate with the platform.
+     * @return a String
+     */
+    String getType();
```
marcelhollerbach, line 37: isnt there some fancy way to mark something "final for the class?"

LeanderK, line 37: no, unfortunately not

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/crowdworking/FallbackWorker.java
```diff
@@ -0,0 +1,35 @@
+package edu.kit.ipd.crowdcontrol.objectservice.crowdworking;
```
marcelhollerbach: Does that really belong in the crowdworking package ? I mean its the rest api which defines that its the field email

marcelhollerbach: Does that really belong in the crowdworking package ? I mean its the rest api which defines that its the field email

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/crowdworking/Platform.java
marcelhollerbach, line 37: :(

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/crowdworking/PlatformManager.java
```diff
@@ -52,12 +69,13 @@ public PlatformManager(List<Platform> crowdPlatforms, WorkerIdentification fallb
 
         //create hashmap of platforms
         platforms = crowdPlatforms.stream()
-                .collect(Collectors.toMap(Platform::getName, Function.identity()));
+                .collect(Collectors.toMap(Platform::getType, Function.identity()));
```
marcelhollerbach, line 55: this breaks the manager! Assume two mturk platforms one productive one sandbox, the manager operations will now be executed on the same interface instance, ( I think the first which appears) but should not.

LeanderK, line 55: these are the same type? not `mturk` and `murk_sandbox`?

marcelhollerbach, line 55: Sure!

LeanderK, line 55: but thats a problem! I need a non-changing id which can use to identify a platform. when i call `/workers/:platform/identity` i need a string `:platform` which i can use in the worker-service to call the platform. Then the name type is not right.

marcelhollerbach, line 55: I never knew about that, add another map with the type as key, and use it in identify worker (we have to write somewhere that the idenfitifaction should not differ in 2 instances...

kelunik, line 55: It must be possible that the identification differs for two platforms of the same type. Could also be two totallly unrelated Pybossa instances.

PR #87, opened 2016-02-05
crowdWorking: Type and dummy-platform

LeanderK:
- added type to crowd working platform
- added dummy-platform


src/main/java/edu/kit/ipd/crowdcontrol/objectservice/crowdworking/Platform.java
```diff
@@ -11,33 +11,42 @@
...
     String getName();
 
     /**
+     * Returns the type of the platform.
+     * <p>
+     * The String must match {@code [a-z0-9_]+} and must be a constant since it is the only way other services can
+     * communicate with the platform.
+     * @return a String
+     */
+    String getType();
```
marcelhollerbach: is the type now a "type" like mturk or something like a "id" like mturksandbox ?

PR #92, opened 2016-02-05
If Marcellus Wallace demands, you follow

lckr:
Replaced implementation of Observer with lambda-function


marcelhollerbach:
I do not demand! You scrued up the subscribe! Nice try ;)

PR #95, opened 2016-02-05
Crowdplatform db fix

marcelhollerbach:
We have to keep the database consistent, for the case a platform disappears in our installation we need to mark this platform somehow as dead. So we can keep our database halfway consistent.

### DIFF
--- a/src/main/java/edu/kit/ipd/crowdcontrol/objectservice/crowdworking/PlatformManager.java
+++ b/src/main/java/edu/kit/ipd/crowdcontrol/objectservice/crowdworking/PlatformManager.java
@@ -54,20 +54,27 @@ public class PlatformManager {
         //create hashmap of platforms
         platforms = crowdPlatforms.stream()
                 .collect(Collectors.toMap(Platform::getID, Function.identity()));
-        //clear database
-        platformOps.deleteAllPlatforms();
+
         //update database
         platforms.forEach((s, platform) -> {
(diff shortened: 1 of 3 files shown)


marcelhollerbach:
We need a additional field in platforms-table where we can store the "lock" bit which marks a database as dead.

LeanderK:
what is the expected behaviour of the /platforms call? Does it include the inactive platforms?

marcelhollerbach:
added the docs for it ... sorry I forgot them.

LeanderK:
I'll reopen it when i'm finished

LeanderK:
i meant the rest GET /platforms call, inactive is currently to my knowledge not defined in the protobufs, so should i filter them out?

marcelhollerbach:
oh good question @kelunik what do you mean?

kelunik:
What are the implications of a `dead` platform?

marcelhollerbach:
platform which is not configured anymore...

kelunik:
If we still want to show them in old experiments which used those platforms, we have to expose them in `GET /platforms` and add the inactive flag to the proto spec.

marcelhollerbach:
yeah, thats the point, what do you want?

kelunik:
Exposing them with an inactive flag is fine.

marcelhollerbach:
Can you add them to proto ?

kelunik:
Done via https://github.com/coolcrowd/object-service/pull/100.

PR #99, opened 2016-02-06
Integration tests

kelunik:
First version for templates with working coverage.

### DIFF
--- a/src/main/resources/config.sample.yml
+++ b/src/main/resources/config.sample.yml
@@ -1,17 +1,19 @@
-
 database:
-# the url where to find the database
+    # the url where to find the database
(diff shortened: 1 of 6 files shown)


kelunik:
Squashed commits… to many with trial and error on Travis.

src/main/resources/config.sample.yml
```diff
@@ -21,14 +23,15 @@ deployment:
 
 # the defined platforms to run
 platforms :
-# type to use
-    - type: mturk/Pybossa/dummy
```
marcelhollerbach: can you leave this in ? or at least write that mturk pybossa dummy are the only valid types?

PR #102, opened 2016-02-06
Fix integration tests and add calibrations to tests

kelunik:
- Combined all rest integration tests into one class.
- Added calibration tests.


marcelhollerbach:
Can you update your pr message? there are also template tests.

kelunik:
There are no new template tests, they're just merged into one class because we need a single Spark instance.

marcelhollerbach:
I dont care if they are new or not, just write it into the message, then we know what happend here and why it happend.

PR #105, opened 2016-02-06
dummy-platform: fixed bug

LeanderK:
nothing fancy


marcelhollerbach:
ups, this one goes on my card.

marcelhollerbach:
Just push if the evil test is fixed ... your single commit looks absolutly fine.

PR #106, opened 2016-02-06
Duplicate

lckr:
- `DuplicateChecker` implemented
- We can now check answers for duplicates (only Strings for now)


LeanderK:
since nobody adopted it -.-

LeanderK:
is one of the methods scalable?

LeanderK:
If i am right, currently everything (even the image data!) gets recomputed when we compare two, it don't think this is a great solution. Can we persist them in the database?

lckr:
A good solution might be storing the hashes of each answer and for the signature of each image, and delete them both on the completion of the experiment.

Do we have an identifier indicating if an answer contains a picture-url or is just a plain string?

lckr:
Do you think that having >>100 answers is a common situation?

LeanderK:
i don't know if it is common, but i think our software must definitely support a larger amount of answers.
From the database-perspective, this is no difficult.

LeanderK:
well, you destroyed something. I would rebase this onto master and then do a force-push

LeanderK:
some tests for image similarly are needed

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/database/operations/AnswerRatingOperations.java
```diff
@@ -429,4 +429,12 @@ public int getNumberOfFinalGoodAnswers(int idExperiment) {
                     ))
         );
     }
+
+    /**
+     * Returns a mapping of answerRecords of an experiment to their hashes used for duplicate detection
+     * Only returns answers which are not considered duplicates
+     */
+    public Map<AnswerRecord,Long> getMapOfHashesOfNonDuplicateAnswers(int expId){
+        return null;
```
lckr: This method needs to implemented.

LeanderK: i don't think this is a good ideas in case we have a larger amount of answers

LeanderK:
please rebase onto branch db-update, then i can concern myself wit the operations

LeanderK:
@lucaskrauss i don't think that reading all the answers into memory is a good idea since this code will be called very often. What are the requirements for finding duplicates based on the hashes? It think getting the duplicates based on the hash is possible to implemented completely with a sql-query

lckr:
Yes, as we already talked about this, I will change that

LeanderK:
do you have specified which operations i have to implement?

lckr:
Both the setHashToAnswer and getDuplicates(expID)  in AnswerRatingOperations need to be implemented. One might consider to move getDuplicates to ExperimentOperations, but because it deals with answers having it in AnswerRatingOps feels good aswell

LeanderK:
well i don't think the signature of getDuplicates is that great. I would give a hash and get all the duplicates for it, so we don't have to compare every answer with every other answer when somebody submitted their answer

lckr:
You're right (again :) ). Do you think this solution is appropriate?

LeanderK:
better ;) i will push my proposal in a few minutes

LeanderK:
depends on #204

PR #108, opened 2016-02-06
Fixes - WIP

LeanderK:
- insert Rating was just plain wrong. I don't know what happened there. The problem is for the real solution i needed a parameter which was not provided, so i had to update the proto-files
- various bug fixes

The worker-service is tested against this branch, so everybody who wants to run the worker-service must use the object service in this branch.

I also forgot to switch to master, so the quality branch is in there.

This is not ready to merge, but important, so i have already opened the pull-request.


src/main/java/edu/kit/ipd/crowdcontrol/objectservice/database/operations/AnswerRatingOperations.java
```diff
@@ -231,82 +243,71 @@ private int getAnswerCount(int workerID) {
...
-            RatingRecord record = DSL.using(conf)
-                    .insertInto(RATING)
-                    .set(ratingRecord)
-                    .returning()
-                    .fetchOne();
+        create.update(RATING)
+                .set(RATING.RATING_, rating.getRating())
+                .set(RATING.FEEDBACK, feedback)
+                .where(RATING.ID_RATING.eq(rating.getRatingId()))
+                .execute();
```
kelunik, line 285: If it updates, then it's no longer an `insertNewRating`?

LeanderK, line 285: its a work in progress...i wanted to get the ws running ;)

kelunik, line 285: WIP on `t - 1 day`, great! :wink:

PR #111, opened 2016-02-07
Add sample service file

kelunik:


object-service.service
```diff
@@ -0,0 +1,14 @@
+[Unit]
+Description=CrowdControl Object-Service
+
+[Service]
+Type=simple
+WorkingDirectory=/git/crowdcontrol/object-service
+ExecStart=/git/crowdcontrol/object-service/gradlew run
+Restart=always
+RestartSec=1
```
marcelhollerbach: I think this is dangerous, I am not sure if the vm waits for threads to finish with sigterm.
If not you are going to kill all running requests after 1 sec. Which is pretty dangerous for all platform operations, I would set it to 60 or 120, so we can make sure that we dont kill some important requests like publishing or some payment action.

kelunik: Adjusted.

PR #122, opened 2016-02-09
database: fixed bug

LeanderK:
- time was in milliseconds and promo expected seconds


LeanderK:
@marcelhollerbach noted that milliseconds worked when he tested it. @kelunik is it possible that the specification is wrong and you use milliseconds too?

kelunik:
It's a bug in both implementations, not the spec. ;-)

LeanderK:
so what gets altered? the implementation or the spec? If its a bug you can merge the pull request ;)

kelunik:
Implementations. ;-)

PR #124, opened 2016-02-10
currency exchange rate gets now involved in moneytransfer

TheRittler:


src/main/java/edu/kit/ipd/crowdcontrol/objectservice/moneytransfer/MoneyTransferManager.java
```diff
@@ -359,4 +367,25 @@ private String loadMessage(String path) throws MoneyTransferException {
...
+            responseBody = httpclient.execute(httpGet, responseHandler);
+            httpclient.close();
+        } catch (IOException e) {
+            LOGGER.error("Fetching currency exchange rates failed.");
+            throw new MoneyTransferException(e.getMessage());
+        }
+
+        LOGGER.trace("Completed fetching currency exchange rates.");
+
+        return Float.parseFloat(responseBody);
```
kelunik: This _must_ check for bounds like `> 0.5 && < 2` and limit it otherwise.

kelunik:
@LeanderK You had concerns during the last meeting, what's your current position?

@redalertexpert Needs rebase to `master`.

LeanderK:
still the same, it would like the arithmetic money-operations to happen within a well tested library like https://github.com/JodaOrg/joda-money (never used that one). Also i still think we need to save the currency with the amount in the db, since only then we can provide good currency exchange on demand.

LeanderK:
also i would like a speedup, while is it sill laying around with 4 days inactivity @redalertexpert ?

TheRittler:
Doing it tomorrow, I'm also for saving currency together with amounts. But i would prefer a unified api to get the amount, e.g. two methods getAmountInUSD() and getAmountInEUR() for workers and giftcodes. I would say this is the best method to guarantee no mixing up of  € and $

LeanderK:
a library would also prevent mixing up currencies

TheRittler:
that depends on how we save the amount in the database. I don't understand how the library helps us with saving amounts in the db.

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/moneytransfer/MoneyTransferManager.java
```diff
@@ -162,7 +173,7 @@ private void fetchNewGiftCodes() throws MoneyTransferException {
             try {
                 Optional<GiftCodeRecord> rec = MailParser.parseAmazonGiftCode(message, parsingPassword);
                 if (rec.isPresent()) {
-                    workerBalanceOperations.addGiftCode(rec.get().getCode(), rec.get().getAmount());
+                    workerBalanceOperations.addGiftCode(rec.get().getCode(), (int) (rec.get().getAmount() * eurUsdRate));
                     giftCodesCount++;
                 }
```
LeanderK: what happens if it is not present? shouldn't that at least be logged? If we can't parse the email, we are throwing potential money away

LeanderK:
we save the currency with the amount and use a library to do the arithmetic, because money is important and doing things yourself is dangerous, e.g. mixing up currencies, overflow etc.

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/moneytransfer/MoneyTransferManager.java
TheRittler: no, because if it's not present, the mail is no giftcode mail (but e.g. Spam). If the mail is a giftcode mail and cannot be parsed, parseAmazonGiftCode() throws an exception

TheRittler:
I agree with you but i still don't understand how a library prevents us from mixing up currencies. We can just get amounts and currencies, and then use a library to do the arithmetic. What i mean is make it not even possible to just get the amount, and skip library and currencies.

LeanderK:
This is not possible, because there is no datatype money with a currency baked in. One could still do a query and do arithmetic by hand. But good that you agree ;) So lets do it :D We don't have that many days left...

TheRittler:
For this change, i think you have to change some things in the db to add currency to amounts

LeanderK:
yeah, i know -.- Therefore we have to speed up our decision taking. I'll add them to the db, ok?

TheRittler:
Btw: I implemented a platformOperations method on my own, pls look over it!

LeanderK:
the operation looks good

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/moneytransfer/MailParser.java
```diff
@@ -16,6 +16,8 @@
  */
 public class MailParser {
 
+    private static final int currencyCode = 978;
```
kelunik: Which currency is that? Should be more like `CURRENCY_EUR` or so.

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/moneytransfer/MoneyTransferManager.java
```diff
@@ -358,11 +376,65 @@ private String loadMessage(String path) throws MoneyTransferException {
...
+
+        Money money = Money.zero(sourceCurr);
+        money = money.plusMinor(amount);
+        BigDecimal exchangeRate;
+        String key = Integer.toString(sourceCurrencyCode) + Integer.toString(destCurrencyCode);
+        if (exchangeRates.containsKey(key)) {
+            exchangeRate = exchangeRates.get(key);
+        } else {
+            exchangeRate = getExchangeRate(sourceCurr.getCurrencyCode(), destCurr.getCurrencyCode());
+            exchangeRates.put(key, exchangeRate);
```
kelunik, line 436: So this doesn't ever update once it's fetched as long as the software is running stable?

src/test/java/edu/kit/ipd/crowdcontrol/objectservice/moneytransfer/MoneyTransferManagerTest.java
```diff
@@ -42,7 +42,8 @@ public void setUp() throws Exception {
         fetcher = mock(MailFetcher.class);
         payops = mock(WorkerBalanceOperations.class);
         workerops = mock(WorkerOperations.class);
-        mng = new MoneyTransferManager(fetcher, sender, payops, workerops, "pseipd@gmail.com", null, 7, 0);
+        platformOperations = mock(PlatformOperations.class);
+        mng = new MoneyTransferManager(fetcher, sender, payops, workerops, platformOperations, "pseipd@gmail.com", null, 7, 0);
```
kelunik: This should use the main config as well, not something hardcoded.

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/moneytransfer/MoneyTransferManager.java
TheRittler, line 436: no, it does update every time you start submitGiftCodes() (gets resettet at start)

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/moneytransfer/MailParser.java
TheRittler: for sure :D

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/moneytransfer/MoneyTransferManager.java
kelunik, line 436: Fine then.

src/test/java/edu/kit/ipd/crowdcontrol/objectservice/moneytransfer/MoneyTransferManagerTest.java
TheRittler: I don't need this mail address at all, I will delete it.

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/moneytransfer/MoneyTransferManager.java
TheRittler, line 436: oh no, there is a bug: it gets resetted at the end. And when an exception occurs, but some exchange rates got already fetched, the map doesn't get resetted. I'll fix it, and then merge it please ;)

kelunik, line 436: Fixed?

TheRittler, line 436: sure

PR #129, opened 2016-02-11
Crowdplatform code cleanup and fix 

simonkorz:
@fix
errors when connecting to pybossa instance, cause: passing wrong type


src/main/java/edu/kit/ipd/crowdcontrol/objectservice/Main.java
```diff
@@ -79,12 +79,15 @@ else if (config.database.maintainInterval < 0)
...
                     break;
                 case "pybossa":
-                    platformInstance = new PyBossaPlatform(config.deployment.workerService,
+                    PyBossaPlatform pyBossa = new PyBossaPlatform(config.deployment.workerService,
                             platform.apiKey,
                             platform.url,
                             platform.name,
                             platform.projectId,
                             platform.calibrationsAllowed);
+                    pyBossa.init();
```
marcelhollerbach, line 88: why having a seperated init function which is called all the time after the constructor?

marcelhollerbach:
please update

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/Main.java
simonkorz, line 88: it initializes the platform and needs a connection to it. I think you should still be able to construct the platform then

marcelhollerbach, line 88: mhmmm but what is the purpose of it if you have a object in a unusable state if you dont call init before?

kelunik, line 88: Being able to test parts of it without internet.

marcelhollerbach, line 88: how do you want to test pybossa without internet ? :-D

PR #131, opened 2016-02-11
database: fixed bug

LeanderK:
- fixed bug which prohibited submitting multiple ratings per worker

### DIFF
--- a/src/main/java/edu/kit/ipd/crowdcontrol/objectservice/database/operations/AnswerRatingOperations.java
+++ b/src/main/java/edu/kit/ipd/crowdcontrol/objectservice/database/operations/AnswerRatingOperations.java
@@ -2,7 +2,9 @@ package edu.kit.ipd.crowdcontrol.objectservice.database.operations;
 
 import edu.kit.ipd.crowdcontrol.objectservice.database.model.tables.records.*;
 import edu.kit.ipd.crowdcontrol.objectservice.database.transformers.AnswerRatingTransformer;
+import edu.kit.ipd.crowdcontrol.objectservice.database.transformers.ExperimentTransformer;
 import edu.kit.ipd.crowdcontrol.objectservice.proto.CalibrationAnswer;
+import edu.kit.ipd.crowdcontrol.objectservice.proto.Experiment;
 import edu.kit.ipd.crowdcontrol.objectservice.proto.Rating;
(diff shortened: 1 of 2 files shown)


src/main/java/edu/kit/ipd/crowdcontrol/objectservice/database/operations/AnswerRatingOperations.java
```diff
@@ -323,22 +325,25 @@ public Rating updateRating(Rating rating) throws IllegalArgumentException, Illeg
...
-                .map(answerOption -> workerCalibrationOperations
-                                .insertAnswer(workerID, answerOption.getIdCalibrationAnswerOption())
-                );
+                .map(answerOption -> {
+                    try {
+                        return workerCalibrationOperations
+                                        .insertAnswer(workerID, answerOption.getIdCalibrationAnswerOption());
+                    } catch (IllegalArgumentException ignored) {
+                        //is ok, the worker was already added
+                        return CalibrationAnswer.getDefaultInstance();
```
kelunik, line 334: An empty object? Why no just `null`?

LeanderK, line 334: because null and optional don't behave well together

LeanderK, line 334: i don't care whats returned, just that something is

kelunik, line 334: It should probably return the same thing as `workerCalibrationOperatoins.insertAnswer` would return. `CalibrationAnswer.getDefaultInstance` is just an empty object of that time. In that case `null` would just result in a `204 No Content` then.

LeanderK, line 334: i don't return it, it is ok (`private void addToExperimentCalibration`). An empty CalibrationAnswer is valid here

kelunik, line 334: Fine then.

PR #136, opened 2016-02-11
payment information in pybossa tasks

simonkorz:
Adds payment information to task

### DIFF
--- a/src/main/java/edu/kit/ipd/crowdcontrol/objectservice/crowdworking/pybossa/PyBossaPlatform.java
+++ b/src/main/java/edu/kit/ipd/crowdcontrol/objectservice/crowdworking/pybossa/PyBossaPlatform.java
@@ -1,5 +1,6 @@
 package edu.kit.ipd.crowdcontrol.objectservice.crowdworking.pybossa;
 
+import com.google.common.primitives.Ints;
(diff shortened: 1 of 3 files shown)


marcelhollerbach:
pls update

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/crowdworking/pybossa/PyBossaPlatform.java
```diff
@@ -91,7 +91,12 @@ public Boolean isCalibrationAllowed() {
                         .put("expID", experiment.getId())
                         .put("platformName", name)
                         .put("idTasks", new JSONArray(idTasks))
-                        .put("type", "experiment"))
+                        .put("type", "experiment")
+                        .put("paymentBase", experiment.getPaymentBase())
+                        .put("paymentRating", experiment.getPaymentRating())
+                        .put("paymentAnswer", experiment.getPaymentAnswer())
+                        //pybossa doesn't support tags
```
marcelhollerbach, line 95: isnt there any way to get in in the title or descripton or anything like that, maybe with fancy "#" ?

Just think of a user which sets up tags, but a not so usefull title, then pybossa should somehow use tags...

PR #139, opened 2016-02-11
Fix usage of LOGGER.error

kelunik:


src/main/java/edu/kit/ipd/crowdcontrol/objectservice/notification/Notification.java
```diff
@@ -97,7 +97,7 @@ public void run() {
         try {
             policy.invoke(this);
         } catch (RuntimeException e) {
-            LOGGER.error("An Exception occurred while checking or sending a notification.", e);
+            LOGGER.error("An exception occurred while checking or sending a notification.", e);
```
marcelhollerbach, line 100: this messages is like "look here is a error." not needed...

kelunik, line 100: Not related to this PR. This PR just ensures we have stack traces everywhere instead of just the exception name.

marcelhollerbach, line 100: this line printent stack traces before, you just changed Exception to exception ...

kelunik, line 100: Yes, this line is just a spelling fix… But the actual message is unrelated.

marcelhollerbach, line 100: aaah ;) just wondered.

PR #143, opened 2016-02-13
Travis performance

LeanderK:
i had this laying around...
- merged the 2 gradle commands in travis into one
- fixed a bug in build.gradle which tricked grade into generating the wsdl files every time you issue a command (which caused a complete recompile)


build.gradle
```diff
@@ -208,7 +208,7 @@ file("src/main/generated/").mkdirs()
 
 wsdl2java {
     generatedWsdlDir = file("src/main/generated/")  // target directory for generated source coude
```
marcelhollerbach, line 210: you should also alter this one.

PR #146, opened 2016-02-28
Fix workerIdentfication to not call identifyWorker() more than once

simonkorz:
Fix workerIdentfication to not call identifyWorker() more than once per request, as there are platforms that might not be able to identify a worker for given parameters more than once.

### DIFF
--- a/src/main/java/edu/kit/ipd/crowdcontrol/objectservice/crowdworking/PlatformManager.java
+++ b/src/main/java/edu/kit/ipd/crowdcontrol/objectservice/crowdworking/PlatformManager.java
@@ -233,12 +233,12 @@ public class PlatformManager {
     /**
      * Get a worker if he exists
      * @param name Name of the platform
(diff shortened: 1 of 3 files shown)


LeanderK:
this is not how it works....or how it was designed to work. What exactly is your problem?

LeanderK:
also we don't want to rely on string identification. this will change to make things easier (amazon was blocking). Please write me a message, since the functionality was kinda constrained by the db, which will hopefully improve in the qs-phase

simonkorz:
The underlying problem is that the identification part of the pybossa implementation can only identify a worker for given params once (because we need to clear the used identification task to make it reusable). So my intent was to change the usage of the identifyWorker interface to call identify only once, as this came with least changes to other parts.

kelunik:
Identify must be idempotent, so it must be possible to call it how often you want and it must not have any side effects.

PR #145, opened 2016-02-18
DB: restore sanity

LeanderK:
- reworked the while Task/ExperimentsPlatform/ExperimentsCalibration
  - now every Platform associated to an Experiment is in the table Experiments_Platform
  - every record in Experiments_Platform has a status in Experiments_Platform_Status. The possible states are: 'draft', 'failedPublishing', 'running', 'finished', 'stopping', 'stopped'.
  - every record in Experiments_Platform has a mode stored in Experiments_Platform_Mode. The possible modes are: 'answer', 'rating', 'disabled'
  - every change in Experiments_Platform_Status/Experiments_Platform_Mode results in a new entry, so we can trace the history of a experiment (when it got published on which platform, when the mode changed etc.)
  - the current status/mode is for a platform is the last entry. Every entry has a timestamp associated with it
  - the Experiments_Calibrations are now associated to the entries in Experiments_Platform and not to the experiment itself. Thats where they really belong and how it always should have been.
  - there is no Task table anymore, most of the columns moved into the Experiments_Platform, except status which got its own table
- added a version table to the database, the object-service will now refuse to start if the database has a wrong version. An Migration-Strategy can be added in the future, everything is ready for it
- some minor improvements to Answer, which makes it possible to reserver Answers  (not implemented yet)
- some minor improvement to Experiment (now has description and description_raw, also the auto-increment now starts at 1 and increments with 1 to the index)

PR #147, opened 2016-02-28
Docker

LeanderK:
added docker support.
In the future i want to run `docker-compose up` to start the entire backend to debug, but this is only possible after #145 is merged because the object-service and the worker-service are out of sync. Right now it starts an instance of mysql and the object-service. The docker compose file is not meant for production!

all the data is in the /images directory, but the docker file exists at project root because it makes the docker hub integration simpler.

the docker images are located [here](https://hub.docker.com/u/coolcrowd/dashboard/).

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,17 @@
+FROM java:8
+
+COPY src /src
+COPY build.gradle /build.gradle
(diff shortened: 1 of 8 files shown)


LeanderK:
blocked by #148

PR #149, opened 2016-02-28
Pybossa fix

LeanderK:
changed the functionality of `WorkerIdentification` to be able to call it only once per identify request. I think it's also simpler now, while more flexible. This is useful for pybossa.

### DIFF
--- a/src/main/java/edu/kit/ipd/crowdcontrol/objectservice/crowdworking/mturk/MturkPlatform.java
+++ b/src/main/java/edu/kit/ipd/crowdcontrol/objectservice/crowdworking/mturk/MturkPlatform.java
@@ -20,7 +20,7 @@ import java.util.stream.Collectors;
  * @author MarcelHollerbach
  * @version 0.1
  */
-public class MturkPlatform implements Platform,Payment,WorkerIdentification {
+public class MturkPlatform implements Platform,Payment {
     public static final int THIRTY_DAYS = 60 * 60 * 24 * 30;
     public static final int TWO_HOURS = 60 * 60 * 2;
     private final String name;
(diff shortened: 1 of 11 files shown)


src/main/java/edu/kit/ipd/crowdcontrol/objectservice/crowdworking/pybossa/PyBossaPlatform.java
```diff
@@ -134,6 +141,9 @@ public Boolean isCalibrationAllowed() {
      * @throws UnidentifiedWorkerException if the worker cannot be identified
      */
     private String identifyWorker(Map<String, String[]> param) throws UnidentifiedWorkerException {
+        if (param == null) {
+            return "";
```
simonkorz, line 139: this should not return an empty string, but instead throw an UnidentifiedWorkerException

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/crowdworking/mturk/MturkPlatform.java
```diff
@@ -63,8 +63,18 @@ public Boolean isCalibrationAllowed() {
...
+        if (params != null) {
+            String[] workerIdArray = params.get("mTurkWorkerId");
+
+            if (workerIdArray != null && workerIdArray.length > 0) {
+                workerId = workerIdArray[0];
+            } else {
+                throw new UnidentifiedWorkerException("mTurkWorkerId was not set!");
+            }
+        }
+        return Optional.of(WorkerIdentification.findByIdentification(getID(),workerId));
```
simonkorz: returning an empty string will probably cause a database error or an error in worker operations.
rather throw an exception here.

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/crowdworking/pybossa/PyBossaPlatform.java
LeanderK, line 139: the problem is how we know determine whether the platform supports the getWorker method. The Platform-manager calls getWorker(null) and if the result is not empty it assumes the platform supports identifying workers. It is documented in the interface. when the rest-call calls the method the map will be empty, not null.

PR #150, opened 2016-02-29
Fix get answer count

simonkorz:
getAnswerCount() did count a worker's answers in total, but it should rather return the worker's answer count for a specific experiment


TheRittler:
What is that? Did anyone solve this?

TheRittler:
ok nvm...

PR #151, opened 2016-02-29
@Fix getAnswerCount

simonkorz:
getAnswerCount() did count a worker's answers in total, but it should rather return the worker's answer count for a specific experiment


LeanderK:
my fault, weirdly only in os, not ws. i would also stop mentioning François-Xavier Thorns (@Fix), i don't think he likes that.

PR #153, opened 2016-03-01
Fix #126 (check that rating options are present and valid)

### LINKED ISSUE #126: Check that rating options are present
Check that there are multiple rating options in a experiment when it's saved.

kelunik:


src/main/java/edu/kit/ipd/crowdcontrol/objectservice/rest/resources/ExperimentResource.java
```diff
@@ -443,6 +456,10 @@ private Experiment updateExperimentInfoDraftState(int id, Experiment experiment,
             throw new IllegalStateException("When an experiment is running, only the state is allowed to be changed.");
         }
 
+        if (!experiment.getRatingOptionsList().isEmpty()) {
+            assertRatingOptions(experiment);
+        }
+
```
marcelhollerbach: are we currently updating the options if there is a request to do so?

kelunik: Sure. https://github.com/coolcrowd/object-service/blob/fix-126/src/main/java/edu/kit/ipd/crowdcontrol/objectservice/rest/resources/ExperimentResource.java#L518-L520

marcelhollerbach:
I am wondering if this should be a assertion on the table, or at least in the database operation, so noone can add wrong values, in no way.

LeanderK:
it's not possible to check it in the database, since i can only check inserts one by one and no the batch, so i can't check whether there are 2 elements. But it agree that it does not belong in the resource and that it should be moved to the operations-class.

marcelhollerbach:
okay, so move the 0 | 9 assertion into the db and the number assertion into the operation ?

kelunik:
No, it's right where it is. Otherwise the experiment is already created and the options fail. We would have to revert things then etc.

The database shouldn't have any of this application logic at all.

LeanderK:
i disagree, the bound 0-9 is not application logic but property of the data. And that there should be at least 2 in the database is database-related.

kelunik:
Guess we had a different storage backend like just in memory or redis or XY. These checks are needed anyway. They're totally application logic.

LeanderK:
i agree that they should also be in the application since usually that we can provide more sensible error-messages there, but the bound form 0-9 is not application logic. This is the typo of the data (an integer with a constraint from 0-9)

marcelhollerbach:
nah it should all be in the db, but the minimum of 2 rating options cannot be done in the db, so move it into the ops. BUT there should be just one place where the assertions are done, and you should not be able to get arround the assertion.

kelunik:
We choose to do so, we have to do all database operations in a single transaction with a rollback if anything errors. But again, it's application logic, not storage logic. It's the storage's job to persist the options, not to validate them. Validation is application logic and should be handled by the rest API controller.

marcelhollerbach:
so we have application logic which is able to produce storage entrys which are wrong ? Sounds like awesome code.

LeanderK:
this checks the bound of the data, it should either be put into the operations class or the db, but not the resources. Should everyone who calls one of the methods in the operations class check whether the data has a valid format? no, thats the job the the method.

kelunik:
Then we might need another layer for validation, so the operations call the actual storage then. Because if you write all these checks into the operations, they have to be replicated if you add another storage option that's not MySQL.

LeanderK:
thats what jooq is for. I don't think we currently use any function that has no equivalent on the other major databases. I have not tested it but in principle we could switch to postgresql without any or at least with very little modification (but i would guess without any), the exception being the db.sql.

kelunik:
What should we do now? Duplicate the check into the database operations?

marcelhollerbach:
put them in the db operations.

kelunik:
If you explain how we should recover if it fails to insert the options…

marcelhollerbach:
Mhmm "J-U-S-T D-O I-T" should work fine.

Just update the database again and delete the records whichbare there for now.

LeanderK:
what do you mean with recover if it fails to insert the options?

kelunik:
Well… We have that issue anyway if the rating options or tags fail because MySQL is unavailable for a short time, but the experiment record itself is already successfully updated. That's also the case which has to be recovered @LeanderK. Because the REST call should be atomic, no partial update.

marcelhollerbach:
Correct, recovery needs to be done. So no need to discuss the special case rating options ;)

LeanderK:
i tried to retain atomicity while not sacrificing simplicity where possible in the operations-methods. Real atomicity is often not that easy. I think this might be something one can look towards implementing in the future, since then we implemented a sensible, composable transaction mechanism (this is really hard!). We can check basic things in the resource class, but for the rest you have to call the methods. If this causes trouble, we can optimise special operations, but we have to focus somewhere.

kelunik:
But as long as we don't have it, it's way better to have it here.

Actually, the Rest API could even use a different representation than the DB.

LeanderK:
i was talking about atomicity. My whole points is that these checks must be in the db-operations. You can duplicated them if you want in the resources, but they MUST be in the operations class, since other classes could call these methods and their input must be checked as well.

marcelhollerbach:
just handle recovery correctly, checking in the database classes is completly fine.

kelunik:
Database operation checks it now.

marcelhollerbach:
Okay, at least they are also in the database now, but pretty much 1:1 duplicated - BAD!. Why not do recovering if the database operations are failing?

kelunik:
Because it would always create a new experiment first, which would have to be deleted, which might fail again. Failing early prevents all these issues.

marcelhollerbach:
and if something else fails while creating the rating options ? 
And this comment does not address the duplicated code.

kelunik:
Thought about it … _if something else fails_, then it's the database connection. If there is no database connection available, it's very likely that there's also no database connection available to recover and delete the experiment again.

marcelhollerbach:
you cannot know when it comes back, the next db op should wait until the connection is back ( @LeanderK isnt this happening?) .

kelunik:
Up to date with `master` now. @redalertexpert

TheRittler:
@marcelhollerbach do you agree with these changes? I would merge this pull request then

marcelhollerbach:
I do not if there is still duplicated code.

kelunik:
Refactored into separate method.

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/database/operations/ExperimentOperations.java
```diff
@@ -290,4 +289,16 @@ public void storeExperimentsPlatforms(List<String> platforms, int experimentId)
                 .fetch()
                 .map(Record1::value1);
     }
+
```
marcelhollerbach, line 292: you are probly going to kill me now, but can you add javadoc here ? :smile:

marcelhollerbach, line 292: fixed. (dont know why github doesnt hide it)

marcelhollerbach:
@redalertexpert happy with it, you can go for it.

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/database/operations/ExperimentOperations.java
kelunik, line 292: Because the line you commented on wasn't changed.

PR #156, opened 2016-03-03
mailtest fix and splitting mailhandler (branch name is a bit misleading)

TheRittler:
web Mailtest now working again and splitted MailHandler into mailSender and mailFetcher in moneytransfer package

### DIFF
--- a/src/main/resources/config.sample.yml
+++ b/src/main/resources/config.sample.yml
@@ -77,6 +77,9 @@ mail:
     # disabled = true will ignore the mail configuration and print to the command line
     disabled : false
 
+    #specifies the mail address, that gets displayed as sender in moneytransfer mails
+    sendMail: pseipd@gmail.com
+
 moneytransfer:
     # Minimum amount of money in US cent a worker has to earn before he gets paid off
     payOffThreshold: 0
--- a/src/main/resources/config.sample.yml
+++ b/src/main/resources/config.sample.yml
@@ -78,7 +78,7 @@ mail:
     disabled : false
 
     #specifies the mail address, that gets displayed as sender in moneytransfer mails
-    sendMail: pseipd@gmail.com
+    sendsMailsFrom: pseipd@gmail.com
 
 moneytransfer:
     # Minimum amount of money in US cent a worker has to earn before he gets paid off
--- a/.gitignore
+++ b/.gitignore
@@ -97,12 +97,14 @@ gradle-app.setting
 !gradle-wrapper.jar
 
 #integration-test properties
-gmailLogin.properties
-webLogin.properties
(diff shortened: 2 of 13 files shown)


LeanderK:
i would like to see the example properties file for the mail package

LeanderK:
still waiting for an update

TheRittler:
what is your opinion on how i should document that? We should write a documentation.

LeanderK:
what do you want to document?

TheRittler:
The creation and keys/values of the mail config

LeanderK:
make a github wiki-entry (we don't have a uniform format yet) detailing the package and the usage. Also the key/values should be documented.

src/main/resources/config.sample.yml
```diff
@@ -50,8 +53,10 @@ moneytransfer:
     #the password used, to verify a valid amazon giftcode email (enter password as message at amazon homepage, while buying giftcodes
     parsingPassword: test
 
-    #the mail address to send notification mails to
+    #the mail address to send notification mails about moneytransfer issues to
```
marcelhollerbach: what ? isnt this the mail adress for notifications anymore?

src/main/resources/config.sample.yml
```diff
@@ -43,6 +43,9 @@ mail :
     # disabled = true will ignore the mail configuration and print to the command-line
     disabled : false
 
+    #specifies the mail address, that gets displayed as sender in moneytransfer mails
+    sendMail: pseipd@gmail.com
```
marcelhollerbach: maybe use moneytransferMailAdress, since sendmail is often the acronym for the MDA.

kelunik: Better `from` or `sender` or `address`.

src/main/resources/config.sample.yml
TheRittler: not anymore, it has never been. I dont know which mail address the notification package uses. @phimonk

LeanderK: well, we should fix the mess anyway

LeanderK:
You can also update the CommandLineHandler, it has no need to extend MailHandler anymore

TheRittler:
it does run first and then wait, before it started after 7 days, You mean the commandline handler should just implement both interfaces and doesnt extend from MailHandler?

TheRittler:
github wiki entry added, @LeanderK you can look over it, if its that what you wanted.

LeanderK:
please rebase onto master, then it's good to go!

src/main/resources/config.sample.yml
LeanderK: this must be resolved before merge!

PR #165, opened 2016-03-06
Fix escaping html in notifications

### LINKED ISSUE #142: Notifications need escaping
Notifications need to escape `tokens`, otherwise HTML can be inserted into the e-mail by the selected result.

simonkorz:
This resolves #142


LeanderK:
needs to be updated

PR #159, opened 2016-03-04
Crowdplatform Pybossa MegaFix

simonkorz:
This PR brings
- working workerIdentification
- setting/updating the task presenter for the configured project by fetching it from the worker-ui url 
- sample config for a pybossa platform

PR #169, opened 2016-03-07
Mturk fixes

marcelhollerbach:
this renames from bundle.js to worker_ui.js and from CreativeCrowd to WorkerUi,

it also removes a hardcoded link.

### DIFF
--- a/src/main/resources/mturk/worker-ui/mturk.js
+++ b/src/main/resources/mturk/worker-ui/mturk.js
@@ -30,8 +30,6 @@ function initMturk(platformName, workerServiceUrl, experiment) {
         experiment: experiment,
         preview: preview
     });
-    //load initial bits
-    WorkerUI.load();
     WorkerUI.onSubmitAnswer(function (viewData, submittedData) {
         //TODO log this
     });
@@ -41,4 +39,6 @@ function initMturk(platformName, workerServiceUrl, experiment) {
     WorkerUI.onFinished(function (data) {
         sendFinish();
     });
+    //load initial bits
+    WorkerUI.load();
 }
--- a/src/main/resources/mturk/worker-ui/mturk.js
+++ b/src/main/resources/mturk/worker-ui/mturk.js
@@ -21,7 +21,7 @@ function initMturk(platformName, workerServiceUrl, experiment) {
     }
 
     //init lib for workerService
-    CreativeCrowd.init(properties = {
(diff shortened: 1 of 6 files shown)


LeanderK:
just bug-fixes?

marcelhollerbach:
nah well more of b0rk fixes, since the bundle was renamed, and there is also the brand new bling bling feature of NO MORE hardcoded url´s !!!!

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/crowdworking/PlatformManager.java
```diff
@@ -126,12 +128,11 @@ public Payment getPlatformPayment(String name) {
      *
      * @param name The name of the platform
      * @param experiment The experiment to publish
-     * @return None if the platform does not exist
+     * @throws PreActionException if this gets thrown the experiment is not published.
+     * @return A completable future object which indicates if the publishing was successful or not
```
LeanderK: i would still improve that a little bit...like: a normally completed future represents success, an exceptionally completed future represents failure

src/main/resources/mturk/worker-ui/mturk.js
```diff
@@ -31,14 +31,14 @@ function initMturk(platformName, workerServiceUrl, experiment) {
         preview: preview
     });
     //load initial bits
-    CreativeCrowd.load();
-    CreativeCrowd.onSubmitAnswer(function (viewData, submittedData) {
+    WorkerUI.load();
```
simonkorz: WorkerUI.load() should be called after all initialization (e.g. setting callbacks) is done, because they might not be properly registered otherwise

marcelhollerbach: Dude, this is 1:1 the code from you sample. Do you really want to tell me the sample probably doesnt work?

simonkorz: Sorry. I wasn't aware of this problem before, I will update the sample.

marcelhollerbach:
depends on #159

LeanderK:
please update the branch

marcelhollerbach:
Ohm, this branch is up to date :D

marcelhollerbach:
Now it is really up to date.

LeanderK:
ah, beautiful ;)

PR #173, opened 2016-03-08
set experiment to finished if there are no answers to rate for

marcelhollerbach:
Noone will ever get notified for a empty experiment, and noone will ever finish a empty experiment. so do it when calling creative-stopp.


LeanderK:
creative stopped means that nobody can submit new answers anymore. There can still be ratings/answers pending. Ignore the pending answers and put a TODO there, this is an issue that can only resolved with a slight db update (scheduled for today or tomorrow). We still have to wait 2 hours if there are ratings pending.

marcelhollerbach:
Wee need to solve this problem more genericly, this is a issue here and for the ratings.

PR #176, opened 2016-03-09
Mturk showmessages

marcelhollerbach:
- gives the worker a real answer in the statistik.
- remove the useless submit button - I hope we always end with finish, otherwise we are having a ugly problem with mturk.

### DIFF
--- a/src/main/java/edu/kit/ipd/crowdcontrol/objectservice/crowdworking/mturk/MturkPlatform.java
+++ b/src/main/java/edu/kit/ipd/crowdcontrol/objectservice/crowdworking/mturk/MturkPlatform.java
@@ -3,11 +3,16 @@ package edu.kit.ipd.crowdcontrol.objectservice.crowdworking.mturk;
 import com.amazonaws.mturk.requester.doc._2014_08_15.Assignment;
 import com.amazonaws.mturk.requester.doc._2014_08_15.AssignmentStatus;
 import com.amazonaws.mturk.requester.doc._2014_08_15.BonusPayment;
+import com.google.common.base.Charsets;
+import com.google.common.io.CharStreams;
+import edu.kit.ipd.crowdcontrol.objectservice.Main;
 import edu.kit.ipd.crowdcontrol.objectservice.crowdworking.*;
 import edu.kit.ipd.crowdcontrol.objectservice.crowdworking.mturk.command.*;
 import edu.kit.ipd.crowdcontrol.objectservice.proto.Experiment;
 import edu.kit.ipd.crowdcontrol.objectservice.proto.Tag;
 
+import java.io.IOException;
+import java.io.InputStreamReader;
 import java.math.BigDecimal;
 import java.util.*;
 import java.util.concurrent.CompletableFuture;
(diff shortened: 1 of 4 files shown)


LeanderK:
what happens when it doesn't end with finish (should not be happening)

marcelhollerbach:
@LeanderK we would get problems with the worker, since we are forced to give them the possibility to submit the hit, also if this does not happen they wont get money, since there is no assignment we could accept.

src/main/resources/mturk/worker-ui/mturk.js
```diff
@@ -30,11 +54,16 @@ function initMturk(platformName, workerServiceUrl, experiment) {
...
     });
+
     WorkerUI.onSubmitAnswer(function (viewData, submittedData) {
-        //TODO log this
+        console.log("Save answer "+submittedData.answer);
+        creativeAnswers.push(submittedData.answer)
     });
-    WorkerUI.onSubmitCalibration(function (data) {
-        //TODO log this
+    WorkerUI.onSubmitRating(function (data) {
```
simonkorz: should be `function (viewData, submittedData)`. I know it's the same in the sample, but i updated that ;)
viewData will return data used to display the view, submittedData will be what the worker-service gets posted

src/main/resources/mturk/worker-ui/mturk.js
```diff
@@ -30,11 +54,16 @@ function initMturk(platformName, workerServiceUrl, experiment) {
...
+        for (var rating in data) {
+            console.log("Save Rating "+rating.rating+" - "+rating.answer);
```
simonkorz: I'm afraid you forgot about the key iterator thing here.

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/crowdworking/mturk/MturkPlatform.java
```diff
@@ -102,6 +140,18 @@ public Boolean isCalibrationAllowed() {
         return new UnpublishHIT(connection, id);
     }
 
+    public <A> List<A> getFullList(int startIndex, Function<Integer, CompletableFuture<List<A>>> producer) throws ExecutionException, InterruptedException {
```
simonkorz: Please add javadoc here

src/main/resources/mturk/worker-ui/mturk.js
marcelhollerbach: Ohm, no not in this case, i dont know why but it works here :/

src/main/resources/mturk/worker-ui/mturk.js
```diff
@@ -30,11 +54,16 @@ function initMturk(platformName, workerServiceUrl, experiment) {
...
     });
+
     WorkerUI.onSubmitAnswer(function (viewData, submittedData) {
-        //TODO log this
+        console.log("Save answer "+submittedData.answer);
+        creativeAnswers.push(submittedData.answer)
     });
-    WorkerUI.onSubmitCalibration(function (data) {
-        //TODO log this
+    WorkerUI.onSubmitRating(function (function (viewData, submittedData)) {
```
simonkorz: Nice double function ;)

marcelhollerbach: FML

Issue #178, opened 2016-03-09
Error when trying to submit Rating


simonkorz:
When trying to submit the following example rating to the worker-service, 

``` json
{
"rating":0,
"ratingId":8,
"experiment":1,
"answerId":2,
"feedback":"",
"constraints":[1]
}
```

the object-service responds with:

```
MessageOrBuilder must have field set: rating
```

No stack trace availiable.

LeanderK:
i am on it!

kelunik:
Fixed via #182

PR #179, opened 2016-03-09
Proto

LeanderK:
updated the photo-files. The Rating in Rating is now boxed, because ratings of 0 should be allowed.


LeanderK:
don't merge, we have to wait for the ws for this one

kelunik:
We don't need boxing to allow 0 values. We just need it where we have to update another value to a value being 0.

LeanderK:
well, it will fix #178

kelunik:
Removing the assertion here: https://github.com/coolcrowd/object-service/blob/proto/src/main/java/edu/kit/ipd/crowdcontrol/objectservice/database/operations/AnswerRatingOperations.java#L267 also fixes it without any BC impact.

LeanderK:
ok, maybe i had not put enough thought in it ;)

PR #181, opened 2016-03-10
Allow log level specification in YML config, resolves #110

### LINKED ISSUE #110: Add log level to config
Log level for Log4j

kelunik:


simonkorz:
Auto format the whole config file, really?

kelunik:
Autoformat didn't work, manually formatted to two spaces. I can revert that to 4 spaces if necessary.

marcelhollerbach:
Or maybe 3 or 5 ? srly, just leave stuff how it is.

kelunik:
Squashed commits into one.

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/config/Config.java
```diff
@@ -17,17 +20,24 @@
     public Deployment deployment;
 
     /**
-     * all specified platforms in the config
+     * Platform configuration.
```
marcelhollerbach: configuration>s< its a array.

marcelhollerbach: and please no "."

src/main/resources/config.sample.yml
```diff
@@ -45,30 +48,45 @@ platforms :
...
+
       # the id of the pybossa project to use for all experiments
       projectId: 1
+
       # if calibrations are allowed on this platform
       calibrationsAllowed: true
 
 # basic configurations for the mail-module
-mail :
+mail:
```
marcelhollerbach: really, its not about this line, its about all changes to the file, do you really need to force everybodys on your personal prefered style? Can you please stop reformatting like that.

TheRittler: This is a config file, that gets seen by the administrator of crowdControl. I think its important, that at least the config file hast unified formatting and style. The impression of the software would be much better, if style is unified.

marcelhollerbach: 1. do you really want to tell me that a user is impressed if the style is unified ? I guess the administrator will be proud if you managed to configure the software and understand the difference between workerLocalUrl and workerPublicUrl or the fact that notificationsEmail is not used by notifications :>
2. I said its _NOT_ about this line.
3. Yes it should be consistent, is it now: NO, is it after this patch: NO.

TheRittler: no, a user is not impressed if the style is unified, but its a bad impression if it looks like the right hand doesn't know what the left hand does.

marcelhollerbach:
```
#hello
a:b
#next op
a:b
```

```
#hello
a:b

#next op
a:b
```

I dont really care if there is a newline or not. And anyway, if fixing it fix it consistent.

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/config/Config.java
kelunik: Done.

src/main/resources/config.sample.yml
kelunik: I just made it consistent and easy to read. Whitespace is really a good thing for configuration files.

marcelhollerbach: its still not consistent. Whitespaces are a personal style, so we are back at the beginning statement.

kelunik: We had whitespace in front of all other documented options already...

marcelhollerbach: Sorry I meant newlines. 
And no you didnt do it infront of _ALL_ other documented options, see line 36 in the file there is still " platforms :" thats why I say its still _NOT_ consistent.

kelunik: Better now?

marcelhollerbach:
yeah, go for it!

PR #184, opened 2016-03-10
DB schema update

LeanderK:
- adds the Answer_Reservation table where the reservations for the answers get persisted
- adds the Rating_Reservation table where the reservations for the ratings get persisted
  (this will hopefully reduce the null pointers ;) )
- transitions to JSON for platform specific worker_data (column platform_data in table Worker) with an index on indentification
- transitions to JSON for platform specific data in experimentsPlatform (column platform_data) with an index on identification (this allows more flexibel data, for example adding the last time a job was run is now possible)
- answers now have a column to store hashes and a system-response column to store feedback from the system, used for example when a duplicate was recognised 
- Experiments_Platform_Status: renamed stopping to creative_stopping and added shutdown
- Experiments_Platform_Mode: rename column stopgap to mode and rename from disabled to normal
- some adjustments in the crowdworking-package necessary for the new schema


LeanderK:
don't merge yet, we have to wait for the worker-service to update

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/crowdworking/Platform.java
```diff
@@ -38,19 +39,22 @@
...
 
     /**
      * Unpublish the given id from the platform, after this call no worker should be able access the before published experiment.
-     * @param id the id of the before published experiment
+     * @param id the id of ExperimentsPlatform the before published experiment
+     * @param data the stored data
      * @return true on success, false or a exception if it failed
      */
-    CompletableFuture<Boolean> unpublishTask(String id);
+    CompletableFuture<Boolean> unpublishTask(int id, JsonElement data);
```
marcelhollerbach: May I ask why you are passing a id here ? That the platforms can use this id they would need to have a databaseconnection, which doesnt look like a good idea, since all db-ops are done in the Manager class.

marcelhollerbach:
Ok I am fine with the rest.

LeanderK:
coolcrowd/worker-service#21 is ready, but depends on coolcrowd/worker-service#20 and progress on the worker-ui

PR #186, opened 2016-03-11
Add port to config

### LINKED ISSUE #109: Add port to config
Port which Spark uses.

kelunik:
Resolves #109

### DIFF
--- a/src/main/java/edu/kit/ipd/crowdcontrol/objectservice/config/Deployment.java
+++ b/src/main/java/edu/kit/ipd/crowdcontrol/objectservice/config/Deployment.java
@@ -7,7 +7,7 @@ public class Deployment {
     /**
      * Port to listen on
      */
-    public Integer port;
+    public int port = 4567;
     /**
      * Deployment origin of the control-ui
      */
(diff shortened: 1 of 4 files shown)


src/main/java/edu/kit/ipd/crowdcontrol/objectservice/config/Deployment.java
```diff
@@ -5,7 +5,11 @@
  */
 public class Deployment {
     /**
-     * Deployment origin of the control ui.
+     * Port to listen on
+     */
+    public Integer port;
```
marcelhollerbach: mhmm if I remember correctly you can add the default value after this field, if no config value is set for it, the default is used. (and isnt a port a ushort?)

kelunik: There's no `ushort` in Java and Spark wants an `int` anyway.

kelunik: Updated it to specify the default value directly in the config.

PR #192, opened 2016-03-11
Fix rating options in templates

kelunik:
Previously, no rating options with 0 as value could be submitted. Also adds tests for this.


kelunik:
Actually fixes another issue with missing `accept` headers.

simonkorz:
@kelunik please update

kelunik:
Done.

PR #194, opened 2016-03-13
Travis with MySQL 5.7

kelunik:
@LeanderK

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,6 +15,7 @@ cache:
   - "$HOME/.gradle/wrapper/"
 
 before_script:
+- mysql --version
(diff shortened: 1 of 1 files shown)


LeanderK:
mariadb 10.1 is compatible with mysql 5.6, not 5.7. I have an idea for an workaround

kelunik:
> MariaDB 10.1 is the current stable (GA) release of MariaDB. It is built on MariaDB 10.0 with features from MySQL 5.6 & 5.7, and entirely new features not found anywhere else.

According to https://downloads.mariadb.org/

LeanderK:
yeah, but for full support we need 10.2, and thats not released yet

PR #201, opened 2016-03-14
Fix null ratings

kelunik:


LeanderK:
i don't think this is the problem, since i have a 0 and not a NULL in my database, so it got correctly inserted

Issue #191, opened 2016-03-11
Answer without valid ratings - RatingQualityByDistribution


simonkorz:
Trying to submit a rating causes error:

``` java
19:20:33.718 [ratpack-compute-1-8] DEBUG edu.kit.ipd.crowdcontrol.workerservice.objectservice.Communication - Trying to submit rating rating_id: 43
experiment_id: 3
rating: 9
worker: 20
 for worker 20 with route /experiments/3/answers/17/rating
19:20:33.826 [ratpack-compute-1-8] ERROR edu.kit.ipd.crowdcontrol.workerservice.ErrorHandler - an internal error occurred
java.util.concurrent.CompletionException: edu.kit.ipd.crowdcontrol.workerservice.InternalServerErrorException: Cause: Object service responded with: Error! Answer without valid ratings. In class edu.kit.ipd.crowdcontrol.objectservice.quality.ratingQuality.RatingQualityByDistribution
```

lckr:
Should be fixed with the db-update, which avoids submitting null-ratings and thus invokes this exception

LeanderK:
push @lucaskrauss 
no, is not fixed with the db-update since you are handling the proto objects. You can't check whether the rating is set when handling the photo-objects, because a rating of 0 counts as not set.

lckr:
The method which throws the exception doesn't use answers at all. The RatingQualityByDistribution only checks ratings. Which are aquired via answerOperations.getRatingsOfAnswer(answerRecord).

But the error arises from a different source

LeanderK:
i can reproduce it, so it's still here

kelunik:
Might be fixed by #201.

lckr:
I know the source and will fix it! Dont worry

lckr:
Fix: #202

LeanderK:
be happy!

PR #202, opened 2016-03-14
Fix for Answer without valid ratings - RatingQualityByDistribution #191

lckr:
- Now request only answers with present ratings for quality-identificaion. This solves the ratingSubmission error and avoids unnecessary db-ops on answers w/o ratings.
- Additionally removes unnecessary checks of answers and ratings not being affected by a submitted rating

### DIFF
--- a/src/main/java/edu/kit/ipd/crowdcontrol/objectservice/database/operations/AnswerRatingOperations.java
+++ b/src/main/java/edu/kit/ipd/crowdcontrol/objectservice/database/operations/AnswerRatingOperations.java
@@ -463,4 +463,14 @@ public class AnswerRatingOperations extends AbstractOperations {
                         .or(DSL.condition(true))
         );
     }
+
+    /**
+     * Get all answers of the specified experiment for which at least one rating is present.
+     * @param expID id of the experiment
+     * @return all answers of the experiment with one or more ratings
+     */
+    public Result<AnswerRecord> getAnswersWithRatingsOfExperiment(int expID){
+        //TODO
+        return null;
+    }
 }
--- a/src/main/java/edu/kit/ipd/crowdcontrol/objectservice/database/operations/AnswerRatingOperations.java
+++ b/src/main/java/edu/kit/ipd/crowdcontrol/objectservice/database/operations/AnswerRatingOperations.java
@@ -472,39 +472,18 @@ public class AnswerRatingOperations extends AbstractOperations {
                     .fetch();
     }
 
-    /**
-     * Get all answers of the specified experiment for which at least one rating is present.
-     * @param expID id of the experiment
-     * @return all answers of the experiment with one or more ratings
-     */
-    public List<AnswerRecord> getAnswersWithRatingsOfExperiment(int expID){
-        Field<Integer> count = DSL.count(RATING.ID_RATING).as("count");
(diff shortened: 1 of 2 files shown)


src/main/java/edu/kit/ipd/crowdcontrol/objectservice/database/operations/AnswerRatingOperations.java
```diff
@@ -463,4 +463,14 @@ public int getNumberOfFinalGoodAnswers(int idExperiment) {
...
     }
+
+    /**
+     * Get all answers of the specified experiment for which at least one rating is present.
+     * @param expID id of the experiment
+     * @return all answers of the experiment with one or more ratings
+     */
+    public Result<AnswerRecord> getAnswersWithRatingsOfExperiment(int expID){
+        //TODO
+        return null;
```
lckr: @LeanderK This needs to implemented by you. Thanks in advance

LeanderK:
Since we are already improving the quality-package. Every Query i execute has a little overhead..not much, but calling it inside a for() loop is not, especially since this is one of the few pieces of code where performance is not negligible. Also you are comparing everything all the time somebody drops a rating. This will kill us when we have to cope with some load.

LeanderK:
i have added some db-operations that would eliminate some complexity in the operations-class

LeanderK:
You can first get the AnswerRecord and then the related Ratings (the submitted rating should be in there) and then call the methods

LeanderK:
@lucaskrauss out of date

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/quality/QualityIdentificator.java
```diff
@@ -165,43 +168,37 @@ private void checkExpStatus(ExperimentRecord experiment) {
...
+        answerRatingOperations.setQualityToAnswer(answerRecord, result.get(AnswerQualityStrategy.QUALITY));
 
-            // Checks if quality_assured bit can be set.
-            if ((((double) result.get(AnswerQualityStrategy.NUM_OF_RATINGS) / (double) experimentRecord.getRatingsPerAnswer()) >= 0.8)) {
-                answerOperations.setAnswerQualityAssured(answer);
-            }
+        // Checks if quality_assured bit can be set.
+        if ((((double) result.get(AnswerQualityStrategy.NUM_OF_RATINGS) / (double) experimentOperations.getExperiment(answerRecord.getExperiment())
+                .orElseThrow(() -> new IllegalArgumentException("Error, can't find experiment of given experiment-ID: "+answerRecord.getExperiment())).getRatingsPerAnswer()) >= 0.8)) {
+            answerRatingOperations.setAnswerQualityAssured(answerRecord);
```
LeanderK, line 199: why >= 0.8, what does it do?

lckr, line 199: Wasn't the quality_assured_bit supposed to indicate the presence of a certain amount of ratings for an answer?
The 0.8 is just a threshold for that

LeanderK, line 199: ah ok, thats 80 percent of the answers

PR #203, opened 2016-03-14
Mail cleanup

marcelhollerbach:


marcelhollerbach:
This introduces a bit strange calls into the main, I will refactor the main in another pr, dont wanted to do this also in this one, since this would have been very messy

LeanderK:
what about disabling stuff? Is the docker-config ok without modifications?

marcelhollerbach:
havent changed anything on that. should run like before

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/mail/MailReceiver.java
```diff
@@ -14,25 +9,36 @@
...
-        this.auth = auth;
+    public MailReceiver(Protocol protocol, String user, String password, String host, int port) {
+        this.protocol = protocol;
+        this.user = user;
+        this.password = password;
+        this.host = host;
+        this.port = port;
+
+        props = new Properties();
+        props.setProperty("mail.debug", true+"");
```
kelunik, line 40: How about `"true"`?

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/Main.java
```diff
@@ -196,15 +188,16 @@ private static void boot(DatabaseManager databaseManager, List<Platform> platfor
         DatabaseMaintainer maintainer = new DatabaseMaintainer(databaseManager.getContext(), cleanupInterval);
         maintainer.start();
 
-        MailFetcher mailFetcher = getMailFetcher(mailDisabled, sendMail);
-        MailSender mailSender = getMailSender(mailDisabled, sendMail);
+        MailFetcher mailFetcher = getMailFetcher(mailDisabled, getConfig().mail.moneyReceiver);
```
LeanderK: doesn't that produce an NPE with the docker config?

marcelhollerbach: mhmm other words for that is that the config is incomplete. Shouldnt we find some other solution ?

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/mail/MailReceiver.java
marcelhollerbach, line 40: nah I like to be typo-protected.

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/Main.java
LeanderK: well, we have to find some kind of solution ;)

marcelhollerbach: solution: DELIVERED.

PR #212, opened 2016-03-14
Docker config

### LINKED ISSUE #211: docker config is not working
`Exception in thread "main" edu.kit.ipd.crowdcontrol.objectservice.config.ConfigException: WorkerService is not found
objectservice_1 |   at edu.kit.ipd.crowdcontrol.objectservice.ConfigLoader.configValidate(ConfigLoader.java:86)
objectservice_1 |   at edu.kit.ipd.crowdcontrol.objectservice.ConfigLoader.<init>(ConfigLoader.java:57)
objectservice_1 |   at edu.kit.ipd.crowdcontrol.objectservice.Main.main(Main.java:50)`

LeanderK:
fixed #211 and improved the docker-config

### DIFF
--- a/image/conf/config.yml
+++ b/image/conf/config.yml
@@ -46,7 +46,62 @@ platforms :
           calibrationsAllowed: true
 
 moneytransfer:
+    # Minimum amount of money in US cent a worker has to earn before he gets paid off
     payOffThreshold: 0
+
+    # Password used to verify a valid Amazon GiftCode e-mail
+    # enter password as message on the Amazon homepage when buying GiftCodes
     parsingPassword: test
+
+    # E-mail address to send notification mails to
     notificationMailAddress: pseipd@gmail.com
+
+    # Interval in days, when the workers get paid off
     scheduleInterval: 7
+
+# Defined platforms to run
+platforms:
+    # Type to use:
+    # - mturk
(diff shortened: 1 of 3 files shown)


image/conf/config.yml
```diff
@@ -46,7 +46,62 @@ platforms :
...
+      # Name which is used to display
+      name: Sandbox
+
+      # URL to the platform
+      url: http://example.com/
+
+    - type: pybossa
+
+      # Key to access the PyBossa API
+      apiKey: 8ec92fa1-1bd1-42ad-8524-3d2bab4588b1
```
kelunik, line 83: Which key is this?

simonkorz, line 83: you get this key from a user's profile in pybossa. Therefore create a user, log in and go to your profile.
And the key in this config is from the local pybossa container, i.e. can be shared.

kelunik, line 83: > i.e. can be shared

That was the question, thanks.

LeanderK, line 83: i have a docker-postures container on the docker-hub which contains a partly initialised pybossa (the user and the project was added, but not published, because the tasks have to be added by the os first). It is intended for development and therefore can be shared

PR #216, opened 2016-03-15
Mail cleanup

TheRittler:
added logging and some cleaning and optimizing.

### DIFF
--- a/src/main/java/edu/kit/ipd/crowdcontrol/objectservice/mail/MailSender.java
+++ b/src/main/java/edu/kit/ipd/crowdcontrol/objectservice/mail/MailSender.java
@@ -8,7 +8,6 @@ import java.io.UnsupportedEncodingException;
  *
  * @author Felix Rittler
  */
-
 public interface MailSender {
 
     /**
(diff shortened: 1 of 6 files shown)


src/main/java/edu/kit/ipd/crowdcontrol/objectservice/mail/MailSender.java
```diff
@@ -18,5 +17,5 @@
      * @param subject       the subject of the mail
      * @param message       the content of the mail
      */
-    public void sendMail(String recipientMail, String subject, String message) throws MessagingException, UnsupportedEncodingException;
+    void sendMail(String recipientMail, String subject, String message) throws MessagingException, UnsupportedEncodingException;
```
kelunik, line 20: Any reason for this?

TheRittler, line 20: As far as I know modifiers are useless in interfaces.

TheRittler, line 20: A public modifier implicates, that the Method has a modifier, but it hasn't

kelunik, line 20: Oh, it's the interface, fine then.

PR #230, opened 2016-03-16
fix EventManager

marcelhollerbach:


src/main/java/edu/kit/ipd/crowdcontrol/objectservice/Main.java
```diff
@@ -84,19 +86,23 @@ public static void main(String[] args) throws IOException, ConfigException, SQLE
      * @param operationCarrier Databaseoperations to use
      * @param platformManager PlatformManager to use
      * @param experimentOperator the operations to use for starting stopping experiments
+     * @param eventManager
```
kelunik, line 89: Missing Parameter description

marcelhollerbach:
now good?

I have no idea why, but somewhere a EventManager sneaked into a unrelated javadoc

kelunik:
Otherwise looks good. Please merge then.

PR #242, opened 2016-03-17
Algorithm and Platform Tests

kelunik:
We don't need to assert much here, the calls must just succeed and lists must contain at least one value each.


kelunik:
Rebased.

PR #244, opened 2016-03-17
Exclude bindings from coverage data

kelunik:


kelunik:
Invalid, it's not generated.

PR #252, opened 2016-03-17
fix feedback to really load the correct messages

lckr:
for @marcelhollerbach

### DIFF
--- a/src/main/java/edu/kit/ipd/crowdcontrol/objectservice/Utils.java
+++ b/src/main/java/edu/kit/ipd/crowdcontrol/objectservice/Utils.java
@@ -3,6 +3,8 @@ package edu.kit.ipd.crowdcontrol.objectservice;
 import com.google.common.base.Charsets;
 import com.google.common.io.CharStreams;
 import edu.kit.ipd.crowdcontrol.objectservice.Main;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
 
 import java.io.IOException;
 import java.io.InputStreamReader;
@@ -13,6 +15,7 @@ import java.io.InputStreamReader;
  * @author Marcel Hollerbach
  */
 public class Utils {
+    private static final Logger LOGGER = LogManager.getLogger("Utils");
     /**
      * Loads a file located relative to the main class
      * @param file the filename to load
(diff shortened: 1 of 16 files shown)


src/main/java/edu/kit/ipd/crowdcontrol/objectservice/Utils.java
```diff
@@ -0,0 +1,32 @@
...
+     */
+    public static String loadFile(String file) {
+        try {
+            return CharStreams
+                    .toString(new InputStreamReader(
+                            Main.class.getResourceAsStream(file)
+                            , Charsets.UTF_8));
+        } catch (IOException e) {
+            e.printStackTrace();
+            return "";
```
TheRittler: Is this really wanted? I think it would be better if you throw an exception, if the patch is invalid, the file corrupt etc.

marcelhollerbach: I sound log the exception into the logger (2 sec.). And then return ""; the rest of the programm does not have any chance to repaire this...

src/test/resources/feedback/workermessage1.txt
```diff
@@ -10,15 +10,12 @@ Rating:
...
 Rating: 1
-
 Rating:
 
 Feedback: Foobar2
 Quality: 2
 Rating: 2
-
 Additional feedback: awes0me Message
-
```
TheRittler, line 21: Here should be a new line for better layout
[This](http://i.imgur.com/3du4fhN.png) looks a bit strange, I think.

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/Utils.java
marcelhollerbach: Done.

src/test/resources/feedback/workermessage1.txt
marcelhollerbach, line 21: Okay, I just fixed the test so it works. The problem is somewhere else. This is just a fix that the code is running.

TheRittler, line 21: I'd say, functionality should be same or better after changes :stuck_out_tongue_winking_eye: Especially when i don't see, why this change was needed :wink:

marcelhollerbach, line 21: the end of the file was here a "\n" which is not generated by the message. Just run the test and see what is happening. If you want to change the layout, change it, this pr is about fixing the bug that the message cannot be loaded since the loader uses relative paths based on the $PWD before, which does not work in any way on a productive system, since $PWD could be anything. While doing this change this test was fixed. (The mocks of the database are illegal the records will be null => NPE for a null record).

PR #255, opened 2016-03-17
Localplatform

LeanderK:
added a local platform, which is similar to dummy-platform, except it has no payments, so it defaults to the Payment-package.


marcelhollerbach:
You know that this will trigger REAL payment?

lckr:
Thats intended

PR #259, opened 2016-03-18
Fix Range (hasPredecessor / hasSuccessor)

kelunik:
See coolcrowd/control-ui#14.

### DIFF
--- a/src/main/java/edu/kit/ipd/crowdcontrol/objectservice/database/operations/AbstractOperations.java
+++ b/src/main/java/edu/kit/ipd/crowdcontrol/objectservice/database/operations/AbstractOperations.java
@@ -287,7 +287,7 @@ public abstract class AbstractOperations {
         int toSkip = 0;
 
         if (results.isNotEmpty() && results.get(0).getValue(primaryKey).equals(start)) {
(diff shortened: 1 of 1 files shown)


src/main/java/edu/kit/ipd/crowdcontrol/objectservice/database/operations/AbstractOperations.java
```diff
@@ -304,8 +304,18 @@ protected void assertHasField(MessageOrBuilder messageOrBuilder, int... fields)
...
+        boolean hasPredecessors = toSkip == 1;
+        boolean hasSuccessors = results.size() > limit + toSkip;
+
+        if (toSkip == 0) {
+            primaryKeyCondition = next
+                    ? primaryKey.lessThan(start)
+                    : primaryKey.greaterThan(start);
+
+            // FIXME: Ugly hack, otherwise we end up with id >= ? and id < ?
+            Optional<R> before = query.and(false).or(primaryKeyCondition).limit(1).fetchOptional();
```
kelunik, line 316: @LeanderK Would be nice if we could just clone the query passed as parameter.

PR #261, opened 2016-03-18
set state to shutdown

marcelhollerbach:
This fixes handling of states
- platformManager sets the state of platforms to shutdown if the ending was successfully
- the experimentstate is now better mapped to states of a set of platforms
- deprecate stopped form the database and use finished.


LeanderK:
you have to update the tests

PR #279, opened 2016-03-20
Config rework

marcelhollerbach:
reworked the config handling
- new yaml parser which gives better error messages
- remove calibrations allowed, nasty nasty!
- validate platform config better. And warn if there are settings which are not set
- move jwtsecret to deployment

### DIFF
--- a/src/main/java/edu/kit/ipd/crowdcontrol/objectservice/ConfigLoader.java
+++ b/src/main/java/edu/kit/ipd/crowdcontrol/objectservice/ConfigLoader.java
@@ -58,34 +58,9 @@ public class ConfigLoader {
         }
 
 
-        try {
-            config.platforms = Arrays.stream(config.platforms)
-                    .map(platform -> {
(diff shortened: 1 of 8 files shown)


src/main/java/edu/kit/ipd/crowdcontrol/objectservice/ConfigLoader.java
```diff
@@ -56,34 +58,9 @@ public ConfigLoader() throws FileNotFoundException, ConfigException {
...
-                throw e;
-            }
-        }
-
-        if (config.database.url != null) {
-            config.database.url = appendUtf8Settings(config.database.url);
-        }
-
         configValidate(config);
+
```
lckr, line 62: Should we check if the config.database.url is existing first before appending?

marcelhollerbach, line 62: done in configValidate

PR #280, opened 2016-03-20
refactored notification mail address & now sending notifications for …

TheRittler:
…new added giftcodes


src/main/java/edu/kit/ipd/crowdcontrol/objectservice/moneytransfer/MoneyTransferManager.java
```diff
@@ -297,26 +319,44 @@ private void payWorker(WorkerRecord worker, List<GiftCodeRecord> giftCodes) thro
...
+            InputStreamReader reader;
+            switch (level) {
+                case ERROR:
+                    reader = new InputStreamReader(Main.class.getResourceAsStream("/moneytransfer/errorNotification.properties"));
+                    break;
+                case ISSUE:
+                    reader = new InputStreamReader(Main.class.getResourceAsStream("/moneytransfer/issueNotification.properties"));
+                    break;
+                case GIFTCODES_ADDED:
+                    reader  = new InputStreamReader(Main.class.getResourceAsStream("/moneytransfer/giftcodesAddedNotification.properties"));
```
marcelhollerbach: ^ missing break

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/moneytransfer/MoneyTransferManager.java
```diff
@@ -297,26 +319,44 @@ private void payWorker(WorkerRecord worker, List<GiftCodeRecord> giftCodes) thro
...
+            throw new MoneyTransferException("Error while loading properties for a notification.", e);
         }
+        String subject = properties.getProperty("subject");
+        String mail = Utils.loadFile(properties.getProperty("pathToMessage"));
 
-        mail = mail.append(message);
+        Map<String, String> map = new HashMap<>();
+        map.put("content", message);
+        mail = Template.apply(mail, map);
+        System.out.println(mail);
```
marcelhollerbach: öhm, debugging stuff?

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/moneytransfer/MoneyTransferManager.java
```diff
@@ -297,26 +319,44 @@ private void payWorker(WorkerRecord worker, List<GiftCodeRecord> giftCodes) thro
...
         try {
-            if (notificationText.length() != 0) {
-                LOGGER.trace("Started sending a notification about problems with submission of giftcodes.");
-                mailSender.sendMail(notificationMailAddress, subject, mail.toString());
-                LOGGER.trace("Completed sending a notification about problems with submission of giftcodes.");
+            if (mail.length() != 0) {
+                LOGGER.trace("Started sending a notification about " + properties.getProperty("loggerMessage") + ".");
+                mailSender.sendMail(notificationMailAddress, subject, mail);
+                LOGGER.trace("Completed sending a notification about " + properties.getProperty("loggerMessage") + ".");
```
marcelhollerbach: Why would you log this ?

TheRittler: I guess its interesting to know, if an error occurrs with moneytransfer if the user got a notification mail or not or if it was tried to send a mail. I think we can never log to much in moneytransfer.

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/moneytransfer/MoneyTransferManager.java
TheRittler: added

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/moneytransfer/MoneyTransferManager.java
marcelhollerbach: The reason why I am asking is that adding propertie files just for something like this feels like a unusual solution. 
Why not atting something like that:

```
public enum ErrorMessage {
ERROR("Hard mistake happend", "path/to/message", 
NOTIFY("Notification", "blba/bla/bla"),

private final String name;
private final String path;

public ErrorMessage(String displayName, String path) {
...
}
}
```

For more details: http://stackoverflow.com/questions/2457076/can-i-add-a-function-to-enums-in-java

marcelhollerbach:
Okay, I am okay with it, one more question, why have you moved the mailadress to the mail section, is this address used somewhere outside of moneytransfer?

TheRittler:
simon is going to use it for notifications
so please merge fast, he needs the changes :wink:

marcelhollerbach:
mhmm I have to say that I quite dont like that, for what is it used in notifications ?

marcelhollerbach:
Since we are in a rush

kelunik:
@marcelhollerbach That's fine and planned. @phimonk has to send emails to the admin. We still need another config to configure the sender, not just the receiver.

marcelhollerbach:
You wanted to have a receiver ? Then we would need a fetcher/sender configuration not just a single emailadress. Also is it possible to tell the "plan" ?

PR #285, opened 2016-03-21
crowdplatforms: add getLink

marcelhollerbach:
platform can present a link where to find the published experiments

### DIFF
--- a/src/main/java/edu/kit/ipd/crowdcontrol/objectservice/Main.java
+++ b/src/main/java/edu/kit/ipd/crowdcontrol/objectservice/Main.java
@@ -8,7 +8,6 @@ import edu.kit.ipd.crowdcontrol.objectservice.crowdworking.*;
 import edu.kit.ipd.crowdcontrol.objectservice.crowdworking.dummy.DummyPlatform;
 import edu.kit.ipd.crowdcontrol.objectservice.crowdworking.fallback.FallbackWorker;
 import edu.kit.ipd.crowdcontrol.objectservice.crowdworking.local.LocalPlatform;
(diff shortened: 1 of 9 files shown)


kelunik:
I think we should have separate URLs for API and public worker UI. Otherwise it gets really hacky once we have a `url` which should point to a task instead of the entry point. We would also have to remove e.g. the query string etc. when using the same URL.

Something like `publicUrl`. Or maybe `api` and `url` / `link`

marcelhollerbach:
we can add a seperated field, but the only place where it is usefull is mturk, since there are the two url completly unlinked. In pybossa its a bad idea, the entry point is basically URL+"/api" and I guess the tasks can also be found in something like URL+xxx.

So in the end we would configure complete consistent stuff, and just force the user to search for some urls which are just known to devs. 

And even for the case mturk we are having ALWAYS the same service urls so we can just add them to the code, since they will never change and there are just 2, also not adding them to the config eliminates the possibility for missconfigurations.

kelunik:
Could you add it for PyBossa then? Just a simple + `/api` in the constructor and the base URL as public URL for now?

marcelhollerbach:
I have no knowledge about pybossa, I guess we should call @phimonk here!

simonkorz:
I think i can implement this later this day. Just wait a bit :)

simonkorz:
Currently I can't link to an experiment directly from object-service, because i can't translate the experiment id to the task. This might change if I finish some pybossa enhancements in time.

marcelhollerbach:
Okay then we are fine with the current solution ?

marcelhollerbach:
Okay can you merge it ? :)

PR #287, opened 2016-03-21
Enable usage of spaces in platform names

kelunik:


kelunik:
Don't merge yet, need to find an issue with too many `_`.

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/crowdworking/dummy/DummyPlatform.java
```diff
@@ -75,11 +71,7 @@ public String getName() {
      */
     @Override
     public String getID() {
-        String appendix = "";
-        if (name != null) {
-            appendix = name.toLowerCase();
-        }
-        return ID+appendix;
+        return (ID + " " + name).toLowerCase().replaceAll("[^a-z0-9]", "_");
```
marcelhollerbach: Can we move this regax to a more common place, and use everywhere the same?

marcelhollerbach: Or even furthur move it to the interface, so the ID gets computed everytime with this regax

kelunik: We could call it when getting the ID, but that's also in multiple place. We can also add a method to the interface that calls `getId` and applies the regex in the default implementation.

marcelhollerbach: yes that would be my prefered solution, but with a name like getIdLegal or rename the interfacefunction to getIDRaw so everyone knows that this isnt a complet legal id.

kelunik: I think that can be addressed in a separate PR?

marcelhollerbach: I dont think so, since this PR introduces the context which would be changed.

kelunik: Done.

PR #297, opened 2016-03-21
notification: fixed bug

LeanderK:
fixed illegal null in map (caused the mapping to fail)


LeanderK:
not to future self: possible improvements for operation in master!

PR #300, opened 2016-03-21
Experiment recover

kelunik:
@LeanderK We need a way to delete populations for recovery, everything else should be fine.

### DIFF
--- a/src/main/java/edu/kit/ipd/crowdcontrol/objectservice/rest/resources/ExperimentResource.java
+++ b/src/main/java/edu/kit/ipd/crowdcontrol/objectservice/rest/resources/ExperimentResource.java
@@ -154,7 +154,7 @@ public class ExperimentResource {
                         .filter(constraintRecord -> !constraintRecord.getConstraint().isEmpty())
                         .forEach(tagConstraintsOperations::insertConstraint);
             } catch (DataAccessException e) {
-                throw new InternalServerErrorException("Couldn't persist tags.", e);
+                throw new InternalServerErrorException("Couldn't persist constraints.", e);
             }
 
             try {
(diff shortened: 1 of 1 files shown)


src/main/java/edu/kit/ipd/crowdcontrol/objectservice/rest/resources/ExperimentResource.java
```diff
@@ -126,45 +128,80 @@ public Experiment put(Request request, Response response) {
...
-        experimentOperations.storeRatingOptions(experiment.getRatingOptionsList(), id);
+            try {
+                List<ConstraintRecord> constraints = TagConstraintTransformer.getConstraints(experiment, id);
 
-        populationsHelper.storePopulations(id, experiment.getPopulationsList());
+                constraints.stream()
+                        .filter(constraintRecord -> !constraintRecord.getConstraint().isEmpty())
+                        .forEach(tagConstraintsOperations::insertConstraint);
+            } catch (DataAccessException e) {
+                throw new InternalServerErrorException("Couldn't persist tags.", e);
```
marcelhollerbach: can you change this to constraints?

kelunik: Done.

marcelhollerbach:
I think a few methods can be refactored to increase readability. Nested try´s arent really nice to read. Do we need to rush here ?

kelunik:
It's a bug, but no critical bug. Will just result in a incomplete experiment, which is actually fine for DRAFT state.

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/rest/resources/ExperimentResource.java
```diff
@@ -318,6 +355,29 @@ private Experiment updateExperimentInfoDraftState(int id, Experiment experiment,
...
+        } catch (DataAccessException e) {
+            // restore old state
+            updateLinkedExperimentInfo(id, old, experiment, oldRecord);
+
+            throw new InternalServerErrorException("Updating the experiment info failed, restored old state.", e);
+        }
+
+        try {
+            //update the experiment itself
+            experimentOperations.updateExperiment(experimentRecord);
```
marcelhollerbach, line 369: Can you move those two try´s into one ? The exact fact if info or experiment failed can be seen in the bt. So we have duplicated code here.

src/main/java/edu/kit/ipd/crowdcontrol/objectservice/rest/resources/ExperimentResource.java
```diff
@@ -126,45 +128,80 @@ public Experiment put(Request request, Response response) {
...
+        } catch (DataAccessException e) {
+            throw new InternalServerErrorException("Experiment could not be stored.", e);
+        }
 
-        tags.stream()
-                .filter(tagRecord -> !tagRecord.getTag().isEmpty())
-                .forEach(tagConstraintsOperations::insertTag);
+        try {
+            try {
+                List<TagRecord> tags = TagConstraintTransformer.getTags(experiment, id);
```
marcelhollerbach, line 141: I am quite not sure if this should be its own function which creates the info of a expeirment, makes reading easier

kelunik:
Still misses platform recovery by @LeanderK.

LeanderK:
what do you need?

kelunik:
We need to restore platforms from before the change. Currently we just add platforms in `storePopulations`. But in case something failed, I don't think we want it to appear in the database at all.

LeanderK:
well this is more complicated then one would think. I'll look into it when i have time

kelunik:
Yes, that's why it's not done yet. We could make it the last thing we change.

LeanderK:
there are still some bugs left ;)

kelunik:
Hm?

LeanderK:
the multiple-calibrations bug, simon mentioned that he received an invalid reservation error at the demo and when submitting only one rating was not possible. I'll create some tickets

LeanderK:
i may try to do this tomorrow

LeanderK:
obviously this feature was not fully realised, but the progress must not be thrown away. I would create a issue with a short description and a reference to this pull request, so that future work can start here.

PR #306, opened 2016-03-21
Fix hit extender to work around http client bug

kelunik:
See Mashape/unirest-java#92.


marcelhollerbach:
Okay i will merge it, but Plesse have a look what limes you are changing, just inlining stuff which was not before isn't really necessary.

kelunik:
I had to change the line anyway, `l` is not a really good variable name.

marcelhollerbach:
but likly not the context of this pr ;)
I wish you good luck, If you want to change every single variable name which is not awesome.

PR #313, opened 2016-03-22
Mturk hitextender fix

marcelhollerbach:
- Fixes not extending of HITs because of unsupported Seconds
- Makes ExtendHIT work by refactoring the code which creates the ExtendHIT call to mturk

### DIFF
--- a/src/main/java/edu/kit/ipd/crowdcontrol/objectservice/crowdworking/mturk/command/GetHIT.java
+++ b/src/main/java/edu/kit/ipd/crowdcontrol/objectservice/crowdworking/mturk/command/GetHIT.java
@@ -20,7 +20,7 @@ public class GetHIT extends MturkRestCommand<HIT, GetHITResponse> {
      */
     public GetHIT(MTurkConnection con, String id){
         super(con,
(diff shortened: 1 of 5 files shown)


src/main/java/edu/kit/ipd/crowdcontrol/objectservice/crowdworking/mturk/command/GetHIT.java
```diff
@@ -20,7 +20,7 @@
      */
     public GetHIT(MTurkConnection con, String id){
         super(con,
-                "GetHIT","HITDetail","2014-08-15", GetHITResponse.class,
+                "GetHIT", null, "2014-08-15", GetHITResponse.class,
```
LeanderK, line 23: this was intended?

marcelhollerbach, line 23: yes, with the HITDetail group we get all information but no ID, without we are getting all information also the id.

mturk ¯\_(ツ)_/¯

PR #316, opened 2016-03-24
test added. that shows failing of moneytransfer

TheRittler:


kelunik:
Please put it in `integration-test`, it needs a working mail config and can't be run automatically.

kelunik:
This just adds a test, where does it fix anything?

LeanderK:
i haven't looked into the pr, but the title is really confusing

marcelhollerbach:
Sorry but this state is likly to get never in. The test is pending on log output, which is really really really bad.
Also (mt means moneytransfer):
- This is starting the main method just to check for mt running, which is a complete overkill.
- if someone changes the log text your tests are failing, without the actual context beeing changed.
- The test depends on so much more than just mt, if something changes there, the test would also fail if something else changes.
- Its no predictable. If moneytransfer is executed after 20 min. at the first time (perfomance wise configuration wise) then the test would also fail.

marcelhollerbach:
if you disagree just reopen it :)

TheRittler:
The thing is, i need to call main, bc we need a "working" (in quotation marks, bc i think its the reason for the failure of mt) database. The actual failure is, that the moneytransfer never comes back from the addGiftCode() method in the db (what @LeanderK doesn't believe), so i added a test showing this on advice of @kelunik . Tell me a better variant how to show this, and i will do. I just check for response of mt, I just can see one improvement, that I let the test wait until mt starts and then let the timer go down.

marcelhollerbach:
Well its quite easy to check if addGiftCode() returns or not. And also if this is the case, why not just add a integration test which checks if addGiftCode returns or not ?

marcelhollerbach:
Not that I would also not like to see such a test, since its not something which should be tested ... But it would be a better solution than the current solution ...

LeanderK:
this is also really bad...it takes custom configuration to run it, executes a lot of other stuff (which can cause errors) and depends on unspecified behaviour. I really can't run it right now. Why don't you isolate the error a bit more? Does it occur with mocked emails? Or when you only call the query?

LeanderK:
i tried to run it, but failed since i don't know the right money-transfer config. Tests that show that something is failing should isolate the faulty behaviour. You don't have to find the exact error, but create a minimal, reproducible test and open a new pull request.

PR #317, opened 2016-03-25
config: removed auto-reconnect

LeanderK:
auto-reconnect is handled by our pool manager, the driver is just confusing and  reduces performance since it can affect the timeout under some circumstances

### DIFF
--- a/src/main/resources/config.sample.yml
+++ b/src/main/resources/config.sample.yml
@@ -1,7 +1,7 @@
 # Database configuration
 database:
     # URL where to find the database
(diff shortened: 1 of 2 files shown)


src/main/resources/config.sample.yml
```diff
@@ -81,17 +81,7 @@ mail:
...
-
-    moneytransfer:
-        protocol: smtp
-        from: max.example.com
-        auth:
-            server: smtp.example.com
-            port: 456
-            credentials:
-                user: maxMustermann
-                password: SuperSecret
```
LeanderK, line 94: well...this was not intended

LeanderK, line 94: well...this was not intended

PR #319, opened 2016-03-25
db: fixed bug

LeanderK:
this commit fixes a bug in a query:
 Unknown column 'sum' in 'where clause'

### DIFF
--- a/src/main/java/edu/kit/ipd/crowdcontrol/objectservice/database/operations/WorkerOperations.java
+++ b/src/main/java/edu/kit/ipd/crowdcontrol/objectservice/database/operations/WorkerOperations.java
@@ -20,6 +20,7 @@ import java.util.List;
 import java.util.Optional;
 
 import static edu.kit.ipd.crowdcontrol.objectservice.database.model.Tables.*;
+import static org.jooq.impl.DSL.orderBy;
 
 /**
  * Responsible for the operations involving the worker-table.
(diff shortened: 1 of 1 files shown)


kelunik:
You could really start writing your DB fixes as `Fix WorkerOperations.getWorkerWithCreditBalanceGreaterOrEqual` in commit messages.

LeanderK:
well i had like 2 seconds of time, also who reads commit messages? ;)

kelunik:
Everybody using `git blame` and `git log`.

marcelhollerbach:
- git blame shows you commit hashes not the message
- if we start writing the function names we are fixing into the commit messages we are ending up with a full commented diff, one closer look on git log will show you which function was fixed, and the reason for the fix is described in the commit message, so absolutly no reason for blaming the message.

LeanderK:
haha i just wanted to refer to git blame...i think its simple and expresses what i did. When i am using git blame i already know what code i am looking at, so `Fix WorkerOperations.getWorkerWithCreditBalanceGreaterOrEqual`does not make much sense i think. I could add it after a few lines in case some strange person really reads them, but i don't think this is useful, one could just look at the diff. Especially in git blame long titles are confusing i think.

kelunik:
Not `git blame` by default, but in e.g. GitHub's interface. It's also quite handy when reviewing commits when doing a deploy or tagging a new version and writing a summary of fixes, e.g. viewing https://github.com/coolcrowd/object-service/compare/841c83598c014cac6b1e61af3ccf42bc99450de4...master It's fine when looking at the extended message, but it could have been just without the first line, so it would appear directly as `Fixed Unknown column 'sum' in 'where clause'`.

marcelhollerbach:
githubs interfaces != git blame. 

and also if you want to write a summary of fixes you maybe dont want to know that WorkerOperations.getWorkerWithCreditBalanceGreaterOrEqual is fixed, since this is telling nobody anything. Maybe the result of the issue could be written into the message but really NOT the class+function name.

kelunik:
Almost everything is better for an overview than `db: fixed bug` ;-)

marcelhollerbach:
But definitly not "Fixed WorkerOperations.getWorkerWithCreditBalanceGreaterOrEqual".

LeanderK:
i don't think that `db: fixed bug` is bad. when i look at a bug i want to know what was going on in git blame and it servers that purpose. I changed 2 lines, not much to justify, i just fixed a bug. And which method i changed, thats what the diff is for. Also `Fixed WorkerOperations.getWorkerWithCreditBalanceGreaterOrEqual` is really long and i don't understand the benefit over `fixed bug`

PR #322, opened 2016-07-11
Update Main.java

anileinchen:
changed mturk url

### DIFF
--- a/src/main/java/edu/kit/ipd/crowdcontrol/objectservice/Main.java
+++ b/src/main/java/edu/kit/ipd/crowdcontrol/objectservice/Main.java
@@ -268,7 +268,7 @@ public class Main {
                     //FIXME remove the sandbox url - but I am to paranoid
                     platformInstance = mturkPlatform = new MturkPlatform(platform.user,
                             platform.password,
-                            "https://mechanicalturk.sandbox.amazonaws.com/",
+                            "https://mechanicalturk.amazonaws.com/",
                             platform.name,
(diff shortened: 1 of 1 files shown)


src/main/java/edu/kit/ipd/crowdcontrol/objectservice/Main.java
```diff
@@ -268,7 +268,7 @@ private static void initLogLevel(Config config) {
                     //FIXME remove the sandbox url - but I am to paranoid
                     platformInstance = mturkPlatform = new MturkPlatform(platform.user,
                             platform.password,
-                            "https://mechanicalturk.sandbox.amazonaws.com/",
+                            "https://mechanicalturk.amazonaws.com/",
```
marcelhollerbach, line 271: Hello,

i hardcoded this value that noone uses productive mturk by accident. there is platform.url which will take the value form the config file, which doesnt need a hardcoded string at all here. So platforms can still be used for productive OR sandbox.

anileinchen:
Was just a little hack for me to test something. normally you are totally right. I will solve it in an other way.

LeanderK:
hacks are useful for prototyping, but i would prefer to not have them in the master where possible. If you have any questions, just shoot us an email! we are happy to help.

kelunik:
This should just use `platform.url` finally.

LeanderK:
sometimes i wonder whether you are to idealistic for this kind of work ;) Yes it should, but it does not

PR #323, opened 2016-07-11
Update MturkPlatform.java

anileinchen:
allowing calibrations for a test


marcelhollerbach:
uuuhm, as far as i remember the meetings this completly dissagrees with the AGB´s of amazon ? I thought we are not allowed to ask people for private attributes ...

marcelhollerbach:
@LeanderK anything you remember ? I am actually not that sure about it...

LeanderK:
yeah that was the problem

